PR #5, opened 2015-08-03
Refactor to only need one configuration file per major release.

ehelms:


stbenjam:
Can you do katello_23.yml too? Looks good otherwise

ehelms:
@stbenjam updated

stbenjam:
:+1:

PR #7, opened 2015-08-05
Add Github user forks as an option when building release environment

ehelms:


ehelms:
@stbenjam per your request, this will add forks if given the option and automatically create the version branch if it does not already exist

stbenjam:
Cool, looks good! :+1: ACK from me

PR #18, opened 2016-02-15
Downcase commit message before greping git log for commit

cpeters:


ehelms:
Why do we want to downcase?  The messages won't match exactly if we do that?

cpeters:
@ehelms Yeah your right. Not sure why I did this.

PR #19, opened 2016-02-15
Updating cherry_picker for bugzilla use case

cpeters:
(cherry picked from commit 09d29114b50cfeaf5f1ac08e0288a58435584247)

### DIFF
--- a/lib/tool_belt/cherry_picker.rb
+++ b/lib/tool_belt/cherry_picker.rb
@@ -27,7 +27,7 @@ module ToolBelt
         commits = issue['changesets']
 
         commits.each do |commit|
-          if !commit['comments'].start_with?('Merge pull request') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
+          if commit['comments'].downcase.start_with?('fixes') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
             revisions << commit['revision']
           end
         end
--- a/lib/tool_belt/cherry_picker.rb
+++ b/lib/tool_belt/cherry_picker.rb
@@ -48,6 +48,7 @@ module ToolBelt
         'revision' => revision,
         'repository' => find_repository(revision)
       }
+      pick.merge({'bugzilla_id' => issue['bugzilla_id']}) if issue['bugzilla_id']
     end
 
     def ignore?(id)
@@ -92,7 +93,11 @@ module ToolBelt
     end
 
     def log_entry(pick)
-      "#{pick['id']} - #{Time.parse(pick['closed_on'])}: [#{pick['revision']}] #{pick['subject']}"
+      if pick['bugzilla_id']
+        "#{pick['bugzilla_id']} - #{pick['id']} : [#{pick['revision']}] #{pick['subject']}"
+      else
+        "#{pick['id']} - #{Time.parse(pick['closed_on'])}: [#{pick['revision']}] #{pick['subject']}"
+      end
     end
 
     def find_repository(revision)
--- a/lib/tool_belt/cherry_picker.rb
+++ b/lib/tool_belt/cherry_picker.rb
@@ -27,7 +27,7 @@ module ToolBelt
         commits = issue['changesets']
 
         commits.each do |commit|
-          if commit['comments'].downcase.start_with?('fixes') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
+          if commit['comments'].downcase.start_with?('fixes', 'refs') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
             revisions << commit['revision']
           end
         end
--- a/lib/tool_belt/cherry_picker.rb
+++ b/lib/tool_belt/cherry_picker.rb
@@ -30,11 +30,7 @@ module ToolBelt
         commits = issue['changesets']
 
         if commits.empty?
-          entry = { 'closed' => issue['closed_on'],
-                    'redmine' => { 'id' => issue['id'], 'subject' => issue['subject'] },
-                    'bugzilla' => ({ 'id' => issue['custom_fields'].select { |cf| cf['id'] == 6 }.first['value'], 'summary' => 'TBD' } if self.bugzilla),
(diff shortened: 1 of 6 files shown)


lib/tool_belt/cherry_picker.rb
```diff
@@ -27,7 +27,7 @@ def find_cherry_picks(project, release, repo_names)
         commits = issue['changesets']
 
         commits.each do |commit|
-          if !commit['comments'].start_with?('Merge pull request') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
+          if commit['comments'].downcase.start_with?('fixes') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
```
ehelms: What about cases where 'refs' is used? Do we want to omit commits that don't use fixes/refs? This can occur sometimes in the puppet modules for example.

cpeters: @ehelms I didn't realize that some puppet modules don't follow the commit message standard. I changed this logic because NOT 'Merge pull request' would sometimes bring in several commit hashes and only 1 was a valid hash to cherry-pick with; had to manually compare in redmine vs. the log output

ehelms: Valid in terms of a hash we actually want or valid in terms of a real hash? 
What about 'refs' ? We will need/want those included in the output as well.

cpeters: @ehelms So if I create a PR with a Refs because it doesn't actually resolve all of the redmine issue, that PR is merged, then I finish with a Fixes and that PR is merged. Now the redmine issue is closed; I would need to know both commit hashes for bring the necessary changes downstream; right?

ehelms: Correct. Or vice versa, a commit goes in that 'fixes' an issue. But if that issue isn't officially released, then sometimes a PR is done with 'Refs'.

lib/tool_belt/cherry_picker.rb
```diff
@@ -27,7 +27,7 @@ def find_cherry_picks(project, release, repo_names)
         commits = issue['changesets']
 
         commits.each do |commit|
-          if !commit['comments'].start_with?('Merge pull request') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
+          if commit['comments'].downcase.start_with?('fixes', 'refs') && !@release_environment.commit_in_release_branch?(repo_names, commit['comments'])
```
ehelms: Why do we want to be limited and specific here to fixes and refs?

lib/tool_belt/cherry_picker.rb
```diff
@@ -92,7 +93,11 @@ def write_cherry_pick_log(picks, release)
     end
 
     def log_entry(pick)
-      "#{pick['id']} - #{Time.parse(pick['closed_on'])}: [#{pick['revision']}] #{pick['subject']}"
+      if pick['bugzilla_id']
+        "#{pick['bugzilla_id']} - #{pick['id']} : [#{pick['revision']}] #{pick['subject']}"
+      else
+        "#{pick['id']} - #{Time.parse(pick['closed_on'])}: [#{pick['revision']}] #{pick['subject']}"
+      end
```
ehelms: This will change the output based upon if a Redmine is hooked to a bugzilla which would create uneven and varying output. I think it may be better to just add in the bugzilla ID within the entry if its present. This is where the use of ERB templating would come in way handier than my ad-hoc method.

lib/tool_belt/cherry_picker.rb
cpeters: Those should be the only legitimate prefixes to a commit that should be cherry-picked. Right? I thought that was the agreed upon standard; refs if it's still in progress and fixes if it's complete.

lib/tool_belt/cherry_picker.rb
cpeters: From my testing there was never a 'bugzilla_id' field in any of the picks pulled from Redmine. I injected the bugzilla_id myself when I meshed the bugzilla bug with the corresponding redmine issue. Can you provide an example of a redmine issue that has a bugzilla_id that is actually pulled in with the cherry_picker script?

cpeters: This code is not my final solution; this was a minimally invasive way to keep upstream functionality in place and add downstream features.

lib/tool_belt/cherry_picker.rb
ehelms: It should be -- and the only way it should make it into Redmine, however, I don't see a reason to limit this here currently. The previous intent was just to ensure ugly 'Merge pull request' commits got ignored.

lib/tool_belt/cherry_picker.rb
ehelms: If I look at https://github.com/ehelms/tool_belt/pull/19/files#diff-008c904eac6e32bcbddea7437dbb73d8R51 it pulls the bugzilla_id from the issue data, which would imply that some objects when they reach this point in the code may have that populated and some may not (aka issues without bugzilla's attached). Which would result in varried output format.

lib/tool_belt/cherry_picker.rb
cpeters: During my testing I ran into instances where there were commits that had neither refs nor fixes and they appeared to be commits that should not be cherry-picked.

lib/tool_belt/cherry_picker.rb
cpeters: Correct. But the only way the data of an issue has a bugzilla_id is if I actually injected it. The query to redmine never pulls back issues with a bugzilla_id.

```redmine_urls = bugs['result']['bugs'].collect { |bug| {:url => bug['url'], :bugzilla_id => bug['id']} }```

I associated the bugzilla_id with the redmine url myself.

ehelms: Checkout: http://projects.theforeman.org/issues/13166.json

If you look in the custom fields section you see the Bugzilla data that comes from Redmine.

PR #20, opened 2016-02-22
allow for beta tag pulp cloning

jlsherrill:
and other fixes

### DIFF
--- a/lib/tool_belt/pulp_repository_updater.rb
+++ b/lib/tool_belt/pulp_repository_updater.rb
@@ -3,12 +3,13 @@ require File.join(File.dirname(__FILE__), 'systools')
 module ToolBelt
   class PulpRepositoryUpdater
 
-    attr_accessor :katello_version, :pulp_version, :systools
+    attr_accessor :katello_version, :pulp_version, :systools, :beta
 
-    def initialize(katello_version, pulp_version, systools = SysTools.new)
+    def initialize(katello_version, pulp_version, release, systools = SysTools.new)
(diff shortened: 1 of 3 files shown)


lib/tool_belt/pulp_repository_updater.rb
```diff
@@ -34,21 +35,21 @@ def compare_repos(os_version, client = false)
...
         index = File.read('index.html')
         rpms = index.scan(/href=".*.rpm"/).collect { |link| link.split('href=')[1].gsub('"', '') }
         rpms.each do |rpm|
-          syscall("wget #{pulp_repo(os_version)}#{rpm}")
+          @systools.syscall("wget #{pulp_repo(os_version)}#{rpm}")
         end
-        syscall("sudo yum -y install createrepo") unless syscall('rpm -q createrepo')[1]
-        syscall("createrepo .")
+        @systools.syscall("sudo yum -y install createrepo") unless @systools.syscall('rpm -q createrepo')[1]
+        @systools.syscall("createrepo .")
```
ehelms: Should execute be used here instead of syscall to have dry run support?

jlsherrill:
@ehelms updated it

lib/tool_belt/pulp_repository_updater.rb
jlsherrill: Actually it should be used here, as you need to create repo in order to compare the repos

ehelms:
Thanks @jlsherrill

PR #21, opened 2016-02-24
Pull updates via tracking branch

domcleal:
When setup-environment is called again, the repos may already exist
with outdated branches.


ehelms:
Thanks @domcleal !

PR #22, opened 2016-02-24
Check repos out into the name given in the config file

domcleal:
Allows the configured repo name and git-generated repo name to differ,
useful when adding the same repo twice for multiple branches under two
names.


ehelms:
Thanks @domcleal !

PR #25, opened 2016-03-09
Allow ignores list to contain git revisions

domcleal:
The ignores list in configs/ may now be a hash of ticket IDs to a list
of git revisions to permit more precise ignoring. When a ticket contains
multiple revisions from different projects, this prevents each revision
to be ignored without the danger of missing new commits against the
ticket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,10 @@ The release configuration file maintains information about the release including
       - 7423
       - 8265
       - 7815
+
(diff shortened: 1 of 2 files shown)


ehelms:
Thanks @domcleal !

domcleal:
Sure. Did you mean to merge it?

ehelms:
I did mean to, sorry about!

PR #26, opened 2016-03-12
Ensure thirdparty convention is used for all Pulp repos

ehelms:


ehelms:
@jlsherrill @beav Can you double check this logic?

jlsherrill:
Looks good to me ack

PR #28, opened 2016-03-16
update to client tagging

beav:
Previous version would update existing packages for clients but not add new
packages.


ehelms:
Thanks @beav !

PR #30, opened 2016-04-11
Support branches, tags and hashes in configuration for repos

ehelms:


ehelms:
Opening this for review, if none will merge in a few days. This change is to help support things like puppet modules that get tagged but not branched (e.g. https://github.com/ehelms/tool_belt/blob/master/configs/katello_30.yaml#L35)

PR #32, opened 2016-06-07
Check older releases for missing commits

ehelms:


jlsherrill:
having not actually tested,  ack

jlsherrill:
in other words, a QA ack

domcleal:
Works well, thanks!

PR #34, opened 2016-06-23
add fedoras to external repos for koji builder

jlsherrill:


ehelms:
ACK - should we consider adding F24 for 3.1 given it was just released?

PR #37, opened 2016-12-16
add branching support to setup_environments

jlsherrill:


lib/tool_belt/release_environment.rb
```diff
@@ -20,20 +20,53 @@ def setup(args = {})
...
           @systools.execute("git clone #{repo[:repo]} #{name}") if !File.exist?(name.to_s)
-          if github_username
-            Dir.chdir(name.to_s) do
+
+          Dir.chdir(name.to_s) do
+            @systools.execute("git remote set-url origin #{add_username(repo[:repo], github_username)}") if github_username
+
+            if repo[:version_branch]
+              if branch_exists?("origin/#{repo[:branch]}") &&
+                @systools.execute("git checkout #{repo[:branch]}")
```
ehelms: This would checkout the local copy of the branch, if your intent is to checkout the true origin version, maybe explcitily do so? e.g. `git checkout origin/#repo[:branch]`

lib/tool_belt/release_environment.rb
```diff
@@ -20,20 +20,53 @@ def setup(args = {})
       Dir.chdir(release_directory) do
         @repos.each do |name, repo|
           @systools.execute("git clone #{repo[:repo]} #{name}") if !File.exist?(name.to_s)
-          if github_username
-            Dir.chdir(name.to_s) do
+
+          Dir.chdir(name.to_s) do
+            @systools.execute("git remote set-url origin #{add_username(repo[:repo], github_username)}") if github_username
```
ehelms: Do we want origin to point to a users fork?

configs/katello_32.yaml
```diff
@@ -30,6 +32,7 @@
   :katello-packaging:
     :branch: KATELLO-3.2
     :repo: https://github.com/Katello/katello-packaging.git
+    :version_branch: true
```
ehelms: by the name of this config value, I'm not sure exactly what I am defining

lib/tool_belt/release_environment.rb
jlsherrill: it doesn't point to the users fork, its adding their username to the url for the main fork to keep from having to type it in when pushing

configs/katello_32.yaml
ehelms: What is the other kind of branch it could be?

PR #40, opened 2017-02-21
Add ability to generate mash scripts

jlsherrill:
and check in3.3 mash scripts

### DIFF
--- a/lib/tool_belt/mash_script.rb
+++ b/lib/tool_belt/mash_script.rb
@@ -1,0 +1,34 @@
+module ToolBelt
+  class MashScript
+
+    attr_accessor :filename, :tag, :gpgkey
+
+    def initialize(filename, options = {})
+      self.filename = filename
+      options.each {|k,v| instance_variable_set("@#{k}",v)}
+    end
+
+    def write_file
+      File.open(filename, 'w') do |file|
+        file.write(template)
+      end
+    end
+
+    def template
+      <<-EOF
+[#{self.tag}]
+rpm_path = %(arch)s/os/Packages
+repodata_path = %(arch)s/os/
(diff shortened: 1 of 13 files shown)


lib/tool_belt/mash_script.rb
```diff
@@ -0,0 +1,34 @@
...
+source_path = source/SRPMS
+debuginfo = False
+multilib = False
+multilib_method = devel
+tag = #{self.tag}
+inherit = True
+strict_keys = False
+keys = #{self.gpgkey}
+arches = x86_64
+      EOF
```
ehelms: Thoughts on moving this to a stored file and using something like ERB templating?

jlsherrill: Don't really care either way.  Its making it a bit more complex but nicer for development I guess.  I'd say right now I slightly lean against doing it, if we add a second instance of something similar I'd probably be completely in favor of it.  But i'm okay either way

ehelms:
Looks good, the one general comment. Also, this only generates the mash configs right? Not the actual script we run to mash?

lib/tool_belt/mash_script.rb
ehelms: Hrmm, at first I thought this was generating them and deploying them but the idea is to use the CLI to generate them and then commit them here?

ehelms:
@jlsherrill bump

jlsherrill:
Correct, we could add something maybe but we might want to restructure most
of the script to be reusable if we go that route

jlsherrill:
That was thinking, still requires to be copied to /etc/mash

ehelms:
ACK - thanks @jlsherrill

PR #41, opened 2017-03-29
Add 3.4 Katello config

ehelms:


configs/katello_34.yaml
```diff
@@ -0,0 +1,172 @@
+---
+:project: katello
+:github_org: katello
+:prior_releases:
+  :3.3.1:
+    :redmine_release_id: 219
```
jlsherrill: 3.3.1 shouldn't be needed here.

ehelms:
Updated

jlsherrill:
ACK, looks fine ot me

PR #51, opened 2017-09-01
Add 3.5 mash scripts

akofink:


akofink:
Updated with 3.5.0 fingerprint

PR #55, opened 2017-10-10
35 config updates

akofink:


configs/katello_35.yaml
```diff
@@ -24,18 +24,34 @@
...
+    :branch: 6.0.0
+    :repo: https://github.com/Katello/puppet-foreman_proxy_content
+  :puppet-katello:
+    :branch: 7.0.0
+    :repo: https://github.com/Katello/puppet-katello
+  :puppet-candlepin:
+    :branch: 4.0.0
+    :repo: https://github.com/Katello/puppet-candlepin
+  :puppet-certs:
+    :branch: master
```
ehelms: We should set this to the installers locked version 3.0.0 (https://github.com/Katello/katello-installer/blob/KATELLO-3.5/Puppetfile.lock#L13)

akofink:
Let's wait on merge. I have a couple more version numbers to update for RC2.

akofink:
Ok, should be good to go now.

PR #57, opened 2017-10-19
[RFC] Add branch and branch-docs commands

akofink:
Might cause less mistakes.

### DIFF
--- a/lib/tool_belt/commands/branch_command.rb
+++ b/lib/tool_belt/commands/branch_command.rb
@@ -3,10 +3,9 @@ module ToolBelt
     class BranchCommand < Clamp::Command
 
       parameter "config_file", "Release configuration file"
-      option "--version", 'VERSION', "Version to build for"
 
       def execute
-        config = ToolBelt::Config.new(config_file, version)
+        config = ToolBelt::Config.new(config_file)
         if File.exist?("repos/#{config.options.namespace}")
           repositories = ToolBelt::Repository.parse(config)
           repositories.each do |repo|
--- a/lib/tool_belt/commands/branch_command.rb
+++ b/lib/tool_belt/commands/branch_command.rb
@@ -1,0 +1,21 @@
+module ToolBelt
+  module Command
+    class BranchCommand < Clamp::Command
+
+      parameter "config_file", "Release configuration file"
+      option "--version", 'VERSION', "Version to build for"
+
+      def execute
+        config = ToolBelt::Config.new(config_file, version)
+        if File.exist?("repos/#{config.options.namespace}")
+          repositories = ToolBelt::Repository.parse(config)
+          repositories.each do |repo|
+            repo.create_or_checkout_branch
+          end
+        else
+          puts "You must run `setup-environment` before `branch`."
+        end
+      end
+    end
+  end
+end
--- a/lib/tool_belt/commands/branch_docs_command.rb
+++ b/lib/tool_belt/commands/branch_docs_command.rb
@@ -1,0 +1,22 @@
+module ToolBelt
+  module Command
+    class BranchDocsCommand < Clamp::Command
+
+      parameter 'config_file', 'Release configuration file'
+
+      def execute
+        config = ToolBelt::Config.new(config_file)
+
+        unless config.options.project == 'katello'
+          puts "This command currently only supports the katello project."
+        end
+
+        if File.exist?("repos/#{config.options.namespace}")
+          DocsRepository.parse(config).update_for_release
+        else
+          puts "You must run `setup-environment` before `branch-docs`."
+        end
+      end
+    end
+  end
+end
--- a/lib/tool_belt/docs_repository.rb
+++ b/lib/tool_belt/docs_repository.rb
@@ -1,0 +1,85 @@
+require 'git'
+
+module ToolBelt
+  class DocsRepository
+
+    attr_reader :releases, :prior_releases, :namespace, :project
+
+    def initialize(args = {})
+      @releases = args[:releases]
+      @prior_releases = args[:prior_releases]
+      @namespace = args[:namespace]
+      @project = args[:project]
+    end
+
+    def self.parse(config)
+      self.new(
+        releases: parse_releases(config.options.releases),
+        prior_releases: parse_releases(config.options.prior_releases),
+        namespace: config.options.namespace,
+        project: config.options.project,
+      )
+    end
+
+    def update_for_release
+      Dir.chdir("repos/#{namespace}") do
+        clone_docs_repo unless File.exist?(repo_name)
+        if project == 'katello'
+          Dir.chdir("#{repo_name}/plugins/katello") do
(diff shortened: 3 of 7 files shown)


akofink:
I also want to add support for branching the docs.

ehelms:
If you want to go this explicit route, I think you should remove this functionality that was essentially doing this -- https://github.com/ehelms/tool_belt/blob/master/lib/tool_belt/release_environment.rb#L33

lib/tool_belt/repository.rb
```diff
@@ -0,0 +1,50 @@
...
+      @options = options
+    end
+
+    def self.parse(config)
+      config.options.repos.map do |repo|
+        self.new(*repo, namespace: config.options.namespace)
+      end
+    end
+
+    def branch!
```
ehelms: Why an exclamation method here?

lib/tool_belt/repository.rb
```diff
@@ -0,0 +1,50 @@
...
+      git.checkout('origin/master')
+      git.reset_hard('origin/master')
```
ehelms: This assumes every repository uses master. Foreman repositories in some cases use develop for their default branch.

lib/tool_belt/repository.rb
akofink: Mostly because it does potentially scary things like pushing a new branch upstream.

lib/tool_belt/repository.rb
akofink: Do you think we could add a field in the config file?

lib/tool_belt/commands/branch_command.rb
```diff
@@ -0,0 +1,18 @@
+module ToolBelt
+  module Command
+    class BranchCommand < Clamp::Command
+
+      parameter "config_file", "Release configuration file"
+      option "--version", 'VERSION', "Version to build for"
+
+      def execute
```
ehelms: This command assumes you've cloned repositories already with no warning as such.

lib/tool_belt/repository.rb
```diff
@@ -0,0 +1,50 @@
...
+      @version_branch = args[:version_branch]
+      @main = args[:main]
+      @options = options
+    end
+
+    def self.parse(config)
+      config.options.repos.map do |repo|
+        self.new(*repo, namespace: config.options.namespace)
+      end
+    end
```
ehelms: I feel like there is similar code in the release_environment.rb

akofink: It happens in the `SetupEnvironmentCommand`.

akofink:
Great feedback @ehelms! I've updated to address most of the feedback. Still WIP for branching docs.

akofink:
Updated with `branch-docs` command.

lib/tool_belt/docs_repository.rb
```diff
@@ -0,0 +1,87 @@
...
+      'theforeman.org'
+    end
+
+    private def release_xy(release)
+      release[/^\d+\.\d+/]
+    end
+
+    private def last_release
+      @last_release ||= prior_releases.sort do |release1, release2|
+        x1,y1,z1 = release1.split('.').map(&:to_i)
```
ehelms: Why three digits? I thought we represented in docs based on just two.

lib/tool_belt/commands/branch_docs_command.rb
```diff
@@ -0,0 +1,19 @@
+module ToolBelt
+  module Command
+    class BranchDocsCommand < Clamp::Command
+
+      parameter 'config_file', 'Release configuration file'
+      option '--version', 'VERSION', 'Version to build for'
```
ehelms: This should not be necessary or useful for this command.

lib/tool_belt/commands/branch_command.rb
```diff
@@ -0,0 +1,21 @@
+module ToolBelt
+  module Command
+    class BranchCommand < Clamp::Command
+
+      parameter "config_file", "Release configuration file"
+      option "--version", 'VERSION', "Version to build for"
```
ehelms: This should not be necessary or useful for this command.

ehelms:
This is still unaddressed -- https://github.com/ehelms/tool_belt/pull/57#issuecomment-338283614

lib/tool_belt/release_environment.rb
```diff
@@ -24,23 +24,9 @@ def setup(args = {})
           Dir.chdir(name.to_s) do
             @systools.execute("git remote set-url origin #{add_username(repo[:repo], github_username)}") if github_username
 
-            if repo[:branch]
-              if branch_exists?("origin/#{repo[:branch]}") &&
-                @systools.execute("git checkout origin/#{repo[:branch]}")
-              elsif tag_exists?(repo[:branch]) &&
-                @systools.execute("git checkout #{repo[:branch]}")
-              else
```
ehelms: We still want this functionality or else when you setup environments it won't setup the branch properly based on the config.

ehelms: The only part that conflicts with your new commands is the line below this that attempts to create a branch if it does not exist.

tools.rb
```diff
@@ -8,6 +8,8 @@ class MainCommand < Clamp::Command
 
   subcommand "cherry-picks", "Calculate needed cherry picks for a given release configuration", ToolBelt::Command::CherryPickCommand
   subcommand "changelog", "Generate changelog for a given release", ToolBelt::Command::ChangelogCommand
+  subcommand "branch", "Create and push the branch or tag for each repository as specified in the configuration", ToolBelt::Command::BranchCommand
+  subcommand "branch-docs", "Generate docs for a given release", ToolBelt::Command::BranchDocsCommand
```
ehelms: So one issue is this is very katello specific the way the code is written and this tool is used by at least Foreman and Katello for releases.

akofink: Yeah, I realized this during development. I'm not really sure of a good solution there other than checking the project field or having maybe a project specific namespace (i.e. `Katello::BranchDocsCommand`).

akofink: For now I [added a check on project](https://github.com/ehelms/tool_belt/pull/57/files#diff-ff264a62dc178b509ae2f522f3b5c129R27).

lib/tool_belt/commands/branch_docs_command.rb
```diff
@@ -0,0 +1,18 @@
+module ToolBelt
+  module Command
+    class BranchDocsCommand < Clamp::Command
+
+      parameter 'config_file', 'Release configuration file'
+
+      def execute
+        config = ToolBelt::Config.new(config_file)
+
```
ehelms: Maybe have this exit if the project is not katello here to inform other project users?

akofink:
Updated

ehelms:
Can you fix Travis?

akofink:
Whoops! I guess `private def <method>` was a Ruby 2.1 addition. Updated.

PR #59, opened 2017-11-16
Fix git.tags usage in repository.rb

pcreech:
git.tags is an array, not a hash.  This adds a function to search the
array for the specified tag.

### DIFF
--- a/lib/tool_belt/repository.rb
+++ b/lib/tool_belt/repository.rb
@@ -20,13 +20,17 @@ module ToolBelt
       end
     end
 
+    def tag_exists?(tag)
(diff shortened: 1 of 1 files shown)


ehelms:
@akofink please to review

lib/tool_belt/repository.rb
```diff
@@ -20,13 +20,17 @@ def self.parse(config)
       end
     end
 
+    def tag_exists(tag)
```
akofink: nitpick: methods that return true/false should end with a ?. `def tag_exists?(tag)`

pcreech:
@akofink agreed, change made

PR #65, opened 2018-01-17
Sort changelog, omit github link

xprazak2:
Sort the categories in changelog alphabetically by category name, add an option to generate changelog without links to github commits.

### DIFF
--- a/lib/tool_belt/changelog.rb
+++ b/lib/tool_belt/changelog.rb
@@ -4,9 +4,9 @@ require File.join(File.dirname(__FILE__), 'redmine/issue')
 module ToolBelt
   class Changelog
 
-    attr_accessor :issues, :bugs, :features, :release_environment, :config
+    attr_accessor :issues, :bugs, :features, :release_environment, :config, :omit_gh_link
 
-    def initialize(config, release_environment)
+    def initialize(config, release_environment, omit_gh_link)
       self.config = config
       issue_cache = IssueCache.new(config)
 
@@ -14,6 +14,7 @@ module ToolBelt
       self.bugs = {}
       self.issues = issue_cache.load_issues
       self.release_environment = release_environment
+      self.omit_gh_link = omit_gh_link
 
       generate_entries(@issues)
       changelog = format_entries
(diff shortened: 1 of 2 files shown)


lib/tool_belt/changelog.rb
```diff
@@ -35,9 +36,9 @@ def generate_entry(issue)
 
       list_item = title_string
       list_item += issue_string
-      list_item += ", "
 
-      list_item += hash_string(issue)
+      list_item += ", "
+      list_item += hash_string(issue) unless @omit_gh_link
```
akofink: No link also means no revision SHA? Seems like we would just want to omit the link part (the parenthesis).

lib/tool_belt/commands/changelog_command.rb
```diff
@@ -21,7 +22,7 @@ def execute
           config.options.namespace,
           systools
         )
-        ToolBelt::Changelog.new(config.options, release_environment)
+        ToolBelt::Changelog.new(config.options, release_environment, short?)
```
akofink: Why not use `omit_gh_link?`? It's much more descriptive as a variable and a cli option.

xprazak2:
I kept the SHA and renamed the option to `omit-gh-link`

ehelms:
@xprazak2 I don't mind the option being here, but it would help if you expand on what use case someone would want to omit the GH link to the commit?

xprazak2:
What gets created here is almost the same as the release notes page on [theforeman.org](https://github.com/theforeman/theforeman.org/blob/gh-pages/_includes/manuals/1.16/1.2_release_notes.md), minus the links. And when I needed a list of changes for 1.17, I removed the links to have the same look.

I found out later that we have [a script](https://github.com/theforeman/theforeman.org/blob/gh-pages/scripts/release_notes.rb) to generate release notes. So I guess this is not really needed.

Closing, thanks!

ehelms:
@xprazak2 Sad a little bit. I'd hope we could eliminate scripts to center on this one tool so we generate commonly structured release notes and changelogs.

xprazak2:
Maybe having the links to a GH commit in our release log may not be a bad idea. I will revisit this when generating a changelog next time. I would also like to eliminate theforeman.org/scripts or at leas move the useful code here.

PR #66, opened 2018-02-05
Fetch all repo data before searching for cherry-picks

zjhuntin:


lib/tool_belt/release_environment.rb
```diff
@@ -19,7 +19,8 @@ def setup(args = {})
 
       Dir.chdir(release_directory) do
         @repos.each do |name, repo|
-          @systools.execute("git clone #{repo[:repo]} #{name}") if !File.exist?(name.to_s)
+          @systools.execute("git clone #{repo[:repo]} #{name}") unless !File.exist?(name.to_s)
```
ehelms: I feel like this changes the meaning of this statement to clone only if it already exists?

zjhuntin: I misread the !File.exist? thought File.exist?

ehelms: So this should be reverted?

zjhuntin:
@ehelms updated

PR #77, opened 2018-05-23
Remove the Issue Cache

jturel:
Per an IRC discussion with @ehelms it was determined that the issue cache is no longer needed since Redmine became stronger. Here's what that looks like.

I tested the changelog and cherry-picks for katello 3.7 vs master and they came out the same so I don't think I broke anything - but please do test.


lib/tool_belt/redmine/issue.rb
```diff
@@ -15,10 +16,45 @@ def id
...
+      sets = @raw_data['issue']['changesets']
+      return [] if sets.nil?
+      sets
+    end
+
+    # arrange pull request data similar to .changesets
+    def pull_request_commits
+      return [] if pull_requests.nil? || pull_requests.empty?
+
+      client = Octokit::Client.new
```
ehelms: Since you are re-writing this code a bit anyway, want to add in support for GITHUB credentials in case of rate limiting?

jturel:
@ehelms yeah, we need that! Added it

PR #78, opened 2018-05-24
Bumps for Katello 3.7

jturel:
I preemptively bumped puppet-certs. Let me know if we shouldn't do that for some reason

### DIFF
--- a/configs/katello_37.yaml
+++ b/configs/katello_37.yaml
@@ -35,22 +35,22 @@
     :branch: 3.2.1
     :repo: https://github.com/Katello/katello-host-tools.git
   :bastion:
-    :branch: 6.1.10
+    :branch: 6.1.11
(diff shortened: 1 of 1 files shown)


configs/katello_37.yaml
```diff
@@ -35,22 +35,22 @@
...
   :puppet-katello:
-    :branch: 7.2.0
+    :branch: 8.0.1
     :repo: https://github.com/Katello/puppet-katello
   :puppet-candlepin:
     :branch: 5.1.0
     :repo: https://github.com/Katello/puppet-candlepin
   :puppet-certs:
-    :branch: 4.0.1
+    :branch: 4.0.2
```
ehelms: I think this will end up being 4.1.0?

jturel: Yep. Fixed

PR #79, opened 2018-05-30
Initial config for 1.18

xprazak2:


configs/foreman_118.yaml
```diff
@@ -0,0 +1,140 @@
...
+    :redmine_release_id: 296
+  :1.17.1:
+    :redmine_release_id: 297
+:releases:
+  :1.18.0:
+    :redmine_release_id: 330
+:repos:
+  :foreman:
+    :main: true
+    :branch: 1.18-stable
```
ehelms: This is likely a bit confusing, when this is an actual branch the following has to be included:

```
version_branch: true
```

Otherwise it treats it as a tag.

xprazak2:
Updated with `version_branch: true`

configs/foreman_118.yaml
```diff
@@ -0,0 +1,147 @@
...
+      - epel-7
+      - centos-sclo-rh-rhel-7
+      - centos-sclo-sclo-rhel-7
+      - centos-7-server
+      - rhel-7.0-server
+      - tfm-ror51-rhel-7
+  - name: foreman-1.18-nonscl-rhel7
+    based_off: foreman-nightly-nonscl-rhel7
+    helper_tags:
+      foreman-1.18-rhel7-override: null
```
xprazak2: I got:
```
kkoji add-tag-inheritance --priority=0 foreman-1.18-nonscl-rhel7-build foreman-1.18-nonscl-rhel7-override
ERROR: Usage: kkoji add-tag-inheritance [options] tag parent-tag
(Specify the --help global option for a list of other help options)

kkoji: error: Invalid tag: foreman-1.18-nonscl-rhel7-override
```

so maybe this should be `foreman-1.18-nonscl-rhel7-override`?

ehelms: Ahh yup it should

xprazak2:
I moved the config into foreman folder.

PR #80, opened 2018-05-31
Add Fedora 28, Remove Fedora 26

jturel:
Doing this on nightly first then will adjust Katello 3.7


jturel:
@ehelms what we discussed...

PR #83, opened 2018-06-01
Use nightly in nightly client tags

ehelms:


jlsherrill:
+99999999999999999999
ACK pending rename of tags in koji, updating of mash scripts

PR #87, opened 2018-07-10
Move configs into namespaced folder

ehelms:


ekohl:
Does it still make sense to have them prefixed? i.e. would `config/foreman/1.18.yaml` also work?

ehelms:
@ekohl updated

PR #89, opened 2018-07-11
Update release_id to fixed_version_id for removal of Backlogs

GregSutcliffe:
~Don't merge yet.~ Good to go.

This is the core changes to the code to start the review process. I'm going through the configs to update the release_ids to match the versions in Redmine, will update the PR once they're ready. I've done one Foreman and one Katello file for testing.

@ehelms could you check this works as expected?

### DIFF
--- a/lib/tool_belt/redmine/project.rb
+++ b/lib/tool_belt/redmine/project.rb
@@ -28,7 +28,7 @@ module Redmine
     end
 
     def get_issues_for_release(id)
-      get_issues(:release_id => id, :limit => 100, :status_id => "5")
+      get_issues(:fixed_version_id => id, :limit => 100, :status_id => "5")
     end
 
     def trackers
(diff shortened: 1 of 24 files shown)


GregSutcliffe:
Added a second commit with the config file changes in, feel free to squash on merge.

jturel:
I think you can also take [WIP] off of the PR name :)

lib/tool_belt/redmine/project.rb
```diff
@@ -28,7 +28,7 @@ def get_issues(params = {})
     end
 
     def get_issues_for_release(id)
```
ekohl: This now matches the `get_issues_for_version` method. We can probably just remove this and switch all callers. We should also get rid of the release naming in favor of version where it's Redmine terms.

ehelms: Let's do as a follow up issue as I want to make sure use of these configs is accurate.

PR #92, opened 2018-07-17
Add 1.19 config

ekohl:


configs/foreman/1.19.yaml
```diff
@@ -0,0 +1,149 @@
...
+    arches:
+      - x86_64
+    inherits:
+      foreman-1.19-nonscl-rhel7-build:
+        foreman-1.19-nonscl-rhel7-override: 0
+      foreman-1.19-nonscl-rhel7-override:
+        foreman-1.19-nonscl-rhel7: 0
+    external_repos:
+      - epel-7
+      - puppetlabs-pc1-rhel-7
```
ekohl: Nightly is also using pc1 but I'd rather move to puppet 5

configs/foreman/1.19.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - centos-7-server
+      - rhel-7.0-server
```
ekohl: I copied this, but do we still need this? Can't we just use centos-7-server?

configs/foreman/1.19.yaml
```diff
@@ -0,0 +1,149 @@
...
+    inherits:
+      foreman-1.19-rhel7-build:
+        foreman-1.19-rhel7-override: 0
+        foreman-1.19-nonscl-rhel7: 10
+      foreman-1.19-rhel7-override:
+        foreman-1.19-rhel7: 0
+    external_repos:
+      - epel-7
+      - centos-sclo-rh-rhel-7
+      - centos-sclo-sclo-rhel-7
```
ekohl: I think we can actually remove this repo since we only use RH SCLs.

ehelms: True but we should make these changes to nightly before here.

configs/foreman/1.19.yaml
ehelms: Yes -- the main benefit is that it provides redhat-release package that sets the dist. Without this our packages come out as `el7.centos`. We can either drop this and accept that or create our own version of that RPM to set the dist to not include anything. In either case, we should do this in nightly for the next release.

configs/foreman/1.19.yaml
ehelms: Yep, we have a task to do that. I think we should aim for 1.20 for that.

PR #95, opened 2018-07-19
Various updates

ekohl:
See individual commits

### DIFF
--- a/lib/tool_belt/issue_service.rb
+++ b/lib/tool_belt/issue_service.rb
@@ -3,12 +3,12 @@ require File.join(File.dirname(__FILE__), 'redmine/project')
 module ToolBelt
   # Find issues for a particular project & release
   class IssueService
-    attr_accessor :project_name, :release, :redmine_release_id, :prior_releases
+    attr_accessor :project_name, :release, :redmine_version_id, :prior_releases
 
     def initialize(config)
       self.project_name = config.project
       self.release = config.release
-      self.redmine_release_id = config.redmine_release_id
+      self.redmine_version_id = config.redmine_version_id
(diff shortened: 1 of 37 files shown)


lib/tool_belt/issue_service.rb
```diff
@@ -3,28 +3,30 @@
...
 
-    def release_issues(release_id)
-      project.get_issues_for_release(release_id)
+    def release_issues(version_id)
+      project.issues_for_version(version_id)
     end
 
-    def load_issues
-      issues = release_issues(redmine_release_id)
+    def load_issues(include_prior)
```
ehelms: Can we set this to false by default so functions don't have to pass this in?

ehelms:
Can you paste a sample generated changelog snippet?

lib/tool_belt/issue_service.rb
ekohl: I intended to do that, but clearly didn't.

ekohl:
> Can you paste a sample generated changelog snippet?

https://gist.github.com/ekohl/fd786adae13051856927fd2697cfea90 was generated using this code

ekohl:
I do see that the generated github URLs are just wrong if redmine identifier != github identifier but I'm pretty certain that was already a problem in the existing code.

lib/tool_belt/issue_service.rb
ekohl: Question is if it should default to true or false. True would be backwards compatible but a user might not expect this and false might make more sense.

ehelms: We also now have the multiple releases way of saying an issue was delivered which simplify/complicates things as well.

PR #98, opened 2018-07-24
Add sles to 3.8

ehelms:


mash_scripts/3.8.0/katello-3.8-client-fedora27.mash
```diff
@@ -0,0 +1,12 @@
+[katello-3.8-client-fedora27]
+rpm_path = %(arch)s/os/Packages
+repodata_path = %(arch)s/os/
+source_path = source/SRPMS
+debuginfo = False
+multilib = False
+multilib_method = devel
+tag = katello-3.8-client-fedora27
+inherit = True
+strict_keys = False
```
ekohl: Should this be True so we check if it's really signed?

PR #99, opened 2018-07-26
Use namespace for mash_scripts

ehelms:


ehelms:
@ekohl Good catch updated

lib/tool_belt/config.rb
```diff
@@ -64,7 +64,7 @@ def add_redmine_release(configs, version)
 
       configs[:release] = version.nil? ? releases.keys.first.to_s : version.to_s
       configs[:redmine_version_id] = configs[:releases][configs[:release].to_sym][:redmine_version_id]
-      configs[:namespace] = "#{configs[:project]}_#{releases.keys.first}"
+      configs[:namespace] = "#{configs[:project]}/#{releases.keys.first}"
```
ekohl: The builder had a `sort`. Is this something we want to maintain or is not needed anymore?

ehelms: I think this aims to trust the configuration to be correct. I think the previous sort was perhaps a safety precaution but it was never really needed.

PR #101, opened 2018-08-01
Add Foreman client Koji tag definition

ehelms:


configs/foreman/nightly.yaml
```diff
@@ -98,3 +98,271 @@
       foreman-plugins-nightly-rhel7-dist:
         foreman-plugins-nightly-rhel7: 2
         foreman-plugins-nightly-nonscl-rhel7: 0
+  - name: foreman-client-nightly-rhel5
+    based_off: null
+    helper_tags:
```
zjhuntin: What does a helper_tag do? Does it just act as the tag?

configs/foreman/nightly.yaml
```diff
@@ -98,3 +98,271 @@
...
+      foreman-client-nightly-rhel5-override: null
+      foreman-client-nightly-rhel5: null
+    build_target: foreman-client-nightly-rhel5-build
+    arches:
+      - x86_64
+    inherits:
```
zjhuntin: Circular inheritance?

configs/foreman/nightly.yaml
```diff
@@ -98,3 +98,271 @@
...
+      foreman-client-nightly-rhel5-build:
+        foreman-client-nightly-rhel5-override: 0
+      foreman-client-nightly-rhel5-override:
+        foreman-client-nightly-rhel5: 0
+      foreman-client-nightly-rhel5: {}
+    external_repos:
```
zjhuntin: Do these just act as repos it understands are packages it will be drawing from?

configs/foreman/nightly.yaml
```diff
@@ -98,3 +98,271 @@
...
+        - bash
+        - redhat-release-server
+        - redhat-rpm-config
+        - rhpkg-simple
+        - rpm-build
+        - shadow-utils
+  - name: foreman-client-nightly-fedora27
+    based_off: null
+    helper_tags:
+      foreman-client-nightly-client-fedora27: null
```
zjhuntin: Did you also want to include `foreman-client-nightly-fedora27` here?

configs/foreman/nightly.yaml
```diff
@@ -98,3 +98,271 @@
...
+        foreman-client-nightly-rhel6-override: 0
+      foreman-client-nightly-rhel6-override:
+        foreman-client-nightly-rhel6: 0
+      foreman-client-nightly-rhel6: {}
+    external_repos:
+      - rhel-6.6-server
+      - rhel-6.6-server-optional
+      - epel-6
+    build_groups:
+      build:
```
zjhuntin: Are these packages that should be used to build packages for the tags or are they supposed to already be tagged into the tags?

ehelms: Both. You can define buildroot group packages that get injected into the buildroot before any other actions take place. These are often common to every build packages such ass tar or gzip as well as packages that have RPM macros that need to be defined before a package starts to get built.

configs/foreman/nightly.yaml
ehelms: I am guessing you meant did I want an override tag here. That's a fair question. Will add.

ehelms:
@zjhuntin Thanks for the review and questions. Hopefully my answers shed some light on how things work. Let me know if oyu have any others while I fix up the override issue.

PR #104, opened 2018-08-15
Update logic for how package groups are handled

ehelms:


lib/tool_belt/koji_tag.rb
```diff
@@ -21,14 +21,13 @@ def process
...
-          tools.ensure_package_group(build_package_group_source_tag, build_target, 'srpm-build')
+          tools.ensure_package_group(build_target, 'build', :source_tag => build_package_group_source_tag)
+          tools.ensure_package_group(build_target, 'srpm-build', :source_tag => build_package_group_source_tag)
         end
 
         if build_groups
           build_groups.each do |group, packages|
-            tools.create_package_group(build_target, group)
-            tools.add_group_packages(build_target, group, packages)
+            tools.ensure_package_group(build_target, group, packages: packages)
```
evgeni: above you write `:source_tag => build_package_group_source_tag` but here you use `packages: packages`, while this is equivalent in modern ruby, I don't think we should mix styles here?

ehelms: That's fair -- updated

lib/tool_belt/koji_tools.rb
```diff
@@ -31,11 +31,16 @@ def ensure_packages(tag_name, packages)
...
-    def ensure_package_group(source_tag, dest_tag, package_group)
+    def ensure_package_group(dest_tag, package_group, options)
       unless has_package_group?(dest_tag, package_group)
         create_package_group(dest_tag, package_group)
       end
-      packages = packages_for_group(source_tag, package_group)
+      packages = if source_tag = options.fetch(:source_tag, false)
+                   packages_for_group(source_tag, package_group)
+                 else
+                   options.fetch(:packages)
```
komidore64: should `options.fetch(:packages)` also have a fallback default?

ehelms: My original idea was no, in that if you don't at least supply packages this should fail which fetch will. The alternative would be adding an empty set of packages which could be viable. @komidore64 got a preference?

komidore64: yeah that makes sense. i'm okay with it failing if no packages were given.

PR #107, opened 2018-08-16
Add a Foreman procedures

ekohl:


lib/tool_belt/render.rb
```diff
@@ -0,0 +1,10 @@
+require 'erb'
+require 'ostruct'
+
+def render(template, vars)
+  ERB.new(template).result(OpenStruct.new(vars).instance_eval { binding })
+end
+
+def render_file(filename, context)
+  render(File.read(filename), context)
+end
```
ehelms: I'm not a huge fan of unscoped functions. That feels like dirty Ruby compared to an class or module to handle this.

ehelms: There may be value in these in the future inside a Template class or module for other actions.

jturel:
I like where this is going :+1:

ekohl:
No longer WIP. Please have a thorough look.

I did notice that when there are packaging changes the exact instructions change. See the `prelease` macro which isn't present in 1.19 yet for example. In the future I also want to standardize tasks like tagging (https://pypi.org/project/bump2version/ comes to mind but there others). That shouldn't hold back this PR though.

lib/tool_belt/commands/procedure.rb
```diff
@@ -0,0 +1,56 @@
...
+            context = {
+              release: release,
+              develop: bump_last(release),
+            }
+
+            render('branch', context)
+          end
+        end
+
+        subcommand "release", "Start releasing Foreman" do
```
zjhuntin: Having a subcommand that has the same parameter name and is the same parameter name as the previous subcommand is confusing.  Maybe new-release? Or name the parameter version?

lib/tool_belt/commands/procedure.rb
```diff
@@ -0,0 +1,56 @@
+module ToolBelt
+  module Command
+    class ProcedureCommand < Clamp::Command
+      subcommand "foreman", "Foreman procedures" do
+        subcommand "branch", "Start branching Foreman" do
+          parameter "release", "Version that the branch will have - like 1.20"
```
zjhuntin: I don't think this is totally clear. This should be the version that the foreman currently has and then will be bumped? I guess I don't totally understand why the person entering the command doesn't bump it themselves.

lib/tool_belt/commands/procedure.rb
ekohl: You're right so I'm changing it to:
```
full_version: 1.20.0-RC1
debian_full_version: 1.20.0~rc1-1
version: 1.20.0
branch: 1.20
extra: RC1
```

It's still the release procedure but there's no more overlap.

lib/tool_belt/commands/procedure.rb
ekohl: This is the procedure to branch nightly into 1.20 and then bump nightly to 1.21. In theory you could automatically detect it somewhere but I think that's out of scope for now.

ekohl:
Updated. Changed the parameter names and added validation on the format.

ehelms:
@zjhuntin Please review

PR #112, opened 2018-09-07
Add Katello 3.9 configuration

zjhuntin:


configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+  :3.7.1:
+    :redmine_version_id: 355
+:releases:
+  :3.9.0:
+    :redmine_version_id: 899
+:code_name: New England IPA
+:repos:
+  :katello:
+    :main: true
+    :branch: KATELLO-3.8
```
ekohl: 3.9?

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+    :version_branch: true
+  :katello-installer:
+    :branch: KATELLO-3.8
+    :repo: https://github.com/Katello/katello-installer.git
+    :version_branch: true
+  :foreman-packaging:
+    :branch: rpm/1.19
+    :repo: https://github.com/theforeman/foreman-packaging.git
+  :puppet-katello:
+    :branch: 8.0.2
```
ekohl: We don't know yet which versions will be used for 3.9. I'd leave them out until we do.

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+    :branch: 4.2.2
+    :repo: https://github.com/Katello/puppet-qpid
+  :puppet-foreman_proxy_content:
+    :branch: 7.2.0
+    :repo: https://github.com/Katello/puppet-foreman_proxy_content
+  :puppet-service_wait:
+    :branch: 1.0.4
+    :repo: https://github.com/theforeman/puppet-service_wait.git
+:ignores:
+  - 23322
```
ekohl: Ignores already?

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+  :puppet-certs:
+    :branch: 4.2.0
+    :repo: https://github.com/Katello/puppet-certs
+  :puppet-qpid:
+    :branch: 4.2.2
+    :repo: https://github.com/Katello/puppet-qpid
+  :puppet-foreman_proxy_content:
+    :branch: 7.2.0
+    :repo: https://github.com/Katello/puppet-foreman_proxy_content
+  :puppet-service_wait:
```
ekohl: We dropped service_wait

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+        katello-3.9-thirdparty-rhel7: 0
+      katello-3.9-client-rhel7:
+        katello-3.9-rhel7: 0
+        katello-3.9-thirdparty-pulp-rhel7: 2
+    external_repos:
+      - epel-7
+      - centos-sclo-rh-rhel-7
+      - centos-sclo-sclo-rhel-7
+      - centos-7-server
+      - rhel-7.0-server
```
ekohl: I think we're in the process of dropping this.

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+        katello-3.9-rhel7: 0
+      katello-3.9-rhel7:
+        katello-3.9-thirdparty-rhel7: 0
+      katello-3.9-client-rhel7:
+        katello-3.9-rhel7: 0
+        katello-3.9-thirdparty-pulp-rhel7: 2
+    external_repos:
+      - epel-7
+      - centos-sclo-rh-rhel-7
+      - centos-sclo-sclo-rhel-7
```
ekohl: This should be dropped (from nightly as well)

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,270 @@
...
+    :main: true
+    :branch: KATELLO-3.8
+    :repo: https://github.com/Katello/katello.git
+    :version_branch: true
+  :katello-installer:
+    :branch: KATELLO-3.8
+    :repo: https://github.com/Katello/katello-installer.git
+    :version_branch: true
+  :foreman-packaging:
+    :branch: rpm/1.19
```
ekohl: Won't it be based on 1.20?

configs/katello/3.9.yaml
zjhuntin: It's not there yet, so I hadn't put it in.

zjhuntin: I'll take it out for now.

configs/katello/3.9.yaml
zjhuntin: will do

configs/katello/3.9.yaml
zjhuntin: I guess I wasn't sure if ignores are kept for each one, but they probably are not, I'll go ahead and take them out.

configs/katello/3.9.yaml
zjhuntin: I had not looked to see if the branch was made yet, but I had assumed not because foreman and katello were not made. I'll remove for now.

configs/katello/3.9.yaml
ehelms: Yea - you can drop ignores

configs/katello/3.9.yaml
ehelms: Good prep PR -- I think setting it what it *will* be is good (e.g. KATELLO-3.9) vs. removing

configs/katello/3.9.yaml
ehelms: ACK - drop this and add `centos-7-server-updates` above the `centos-7-server`

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,191 @@
...
+      katello-3.9-rhel7:
+        katello-3.9-thirdparty-rhel7: 0
+      katello-3.9-client-rhel7:
+        katello-3.9-rhel7: 0
+        katello-3.9-thirdparty-pulp-rhel7: 2
+    external_repos:
+      - epel-7
+      - centos-sclo-rh-rhel-7
+      - centos-7-server
+      - foreman-rails-1.19-rhel-7
```
ehelms: This will be `1.20`

configs/katello/3.9.yaml
```diff
@@ -0,0 +1,191 @@
...
+        katello-3.9-rhel7: 0
+      katello-3.9-rhel7:
+        katello-3.9-thirdparty-rhel7: 0
+      katello-3.9-client-rhel7:
+        katello-3.9-rhel7: 0
+        katello-3.9-thirdparty-pulp-rhel7: 2
+    external_repos:
+      - epel-7
+      - centos-sclo-rh-rhel-7
+      - centos-7-server
```
ehelms: Add `centos-7-server-updates` above this entry

PR #120, opened 2018-10-18
Add Katello 3.9 mash scripts

ehelms:


ehelms:
@zjhuntin can you review this

mash_scripts/katello/3.9.0/katello-3.9-rhel7.mash
```diff
@@ -0,0 +1,12 @@
+[katello-3.9-rhel7]
+rpm_path = %(arch)s/os/Packages
+repodata_path = %(arch)s/os/
+source_path = source/SRPMS
+debuginfo = False
+multilib = False
+multilib_method = devel
+tag = katello-3.9-rhel7
+inherit = True
+strict_keys = False
```
ekohl: With Foreman we want strict keys to verify the GPG signing process during RCs. Perhaps also an idea for Katello?

ehelms: Its an idea to consider, we avoided signing RCs in the past due to the overhead.

PR #125, opened 2018-11-14
add buildsys-macros to the installed groups of rhel5 client builds

evgeni:


jturel:
This seems to fix the issue that came about in the referenced PR, so ACK from me, from that perspective

PR #140, opened 2019-01-16
Correct Foreman 1.21 rails tags

ekohl:


ehelms:
What is wrong with the tag names? Looks like you just changed the naming scheme vs. correcting an error

ekohl:
Compare:
`foreman-1.21-tfm-ror52-rhel7`
`foreman-rails-1.21-tfm-ror52-rhel7`

The inheritance refers to a non-existing tag. I'm choosing a naming scheme here that's future compatible: if we move to ror53 then we don't need to really care about this in our mock/rel-eng files: it just works if you specify the right version number.

ehelms:
Thats fair, I liked the explicitness but I can live with this.

PR #145, opened 2019-01-21
Add foreman-1.21 mash-scripts

ehelms:


ekohl:
These are generated from configs. What's the benefit of committing generated files?

ehelms:
In case the generation logic changes, storing them in git allows for change tracking.

PR #152, opened 2019-01-28
Update 1.20 mash configs

ekohl:
Sets strict keys to true and generates a foreman-rails mash config.

### DIFF
--- a/configs/foreman/1.20.yaml
+++ b/configs/foreman/1.20.yaml
@@ -7,6 +7,7 @@
 :mash_scripts:
   - foreman-1.20-rhel7-dist
   - foreman-plugins-1.20-rhel7-dist
+  - foreman-rails-1.20-rhel7
   - foreman-client-1.20-rhel7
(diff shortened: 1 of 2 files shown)


ekohl:
This is incorrect: we shouldn't set strict keys on client repos.

mash_scripts/foreman/1.20.0/foreman-plugins-1.20-rhel7-dist.mash
```diff
@@ -7,6 +7,6 @@ multilib = False
 multilib_method = devel
 tag = foreman-plugins-1.20-rhel7-dist
 inherit = True
-strict_keys = False
+strict_keys = True
```
ehelms: This shouldn't be true on plugins?

ekohl: 5e743d1e5a0bc68cdded58a76f4b6f2debc7ca37

PR #153, opened 2019-01-29
make katello-*-build inherit released foreman rails, not nightly

evgeni:


evgeni:
executed all three against koji.katello.org

PR #155, opened 2019-02-14
Update Foreman release procedure

ekohl:
I now have scripts to do these steps so the various indivdual steps are no longer needed.

In practice we only sign "extra" RPMs during the RC stage. Usually only on the first but this is good enough.

### DIFF
--- a/procedures/foreman/release.md.erb
+++ b/procedures/foreman/release.md.erb
@@ -55,30 +55,18 @@
   * [ ] Tag: `git tag -s -m "Release <%= full_version %>" <%= full_version %>`
   * [ ] Push: `git push --follow-tags`
 * [ ] Run the Jenkins [Tarballs Release](https://ci.theforeman.org/job/tarballs-release/) to create tarballs
-* [ ] Verify tarballs are present on downloads.theforeman.org
-* [ ] Download, sign and upload detached signatures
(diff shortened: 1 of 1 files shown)


procedures/foreman/release.md.erb
```diff
@@ -62,23 +62,10 @@ Note: If for some reason there was an issue with the tarballs that required uplo
...
-  * [ ] Set version to `<%= version %>`
-  * [ ] Set prerelease to `<%= extra %>`
-<% unless is_rc %>  * [ ] Reset release to `1`
-<% end %>  * [ ] In each package dir, remove the old tarball, run `spectool -g *.spec` and `git annex add *.tar.bz2`
-  * [ ] Commit with message "Release <%= full_version %>"
-  * [ ] Submit a pull request
-* [ ] Update changelog files for debs:
-  * [ ] `scripts/changelog.rb -v <%= debian_full_version %> -m "<%= full_version %> released" debian/*/*/changelog`
-  * [ ] Submit a pull request
+* [ ] Update [foreman-packaging](https://github.com/theforeman/foreman-packaging) rpm/<%= short_version %> and deb/<%= short_version %> branches
```
ehelms: What does update mean? Is there a more specific instruction?

ekohl: I am struggling with that a bit. I think I'll update https://projects.theforeman.org/projects/foreman/wiki/Release_Process and document the long form there that describes my script.

ekohl: Now that I've opened the repo: it's calling https://github.com/ekohl/theforeman-rel-eng/blob/master/bump_deb_packaging and https://github.com/ekohl/theforeman-rel-eng/blob/master/bump_rpm_packaging

ekohl:
I've updated it so it's actually the steps I execute.

PR #158, opened 2019-03-13
Move the Foreman nightly build env to Puppet 6

ekohl:
This will need work to actually create this tag before it's merged. I'm also not sure if we should be building with the latest or the oldest supported version.


ehelms:
The external repository has been added

PR #163, opened 2019-04-23
Update the GPG section of Foreman branching

ekohl:


procedures/foreman/branch.md.erb
```diff
@@ -6,10 +6,10 @@
...
 * [ ] Add new languages that are at a [reasonable completion on Transifex](https://www.transifex.com/foreman/foreman/foreman/) to **develop**
-* [ ] Generate a release GPG key using [GPG_Keys](https://projects.theforeman.org/projects/foreman/wiki/GPG_Keys)
- * [ ] publish to keyserver
- * [ ] [security.md](https://github.com/theforeman/theforeman.org/blob/gh-pages/security.md) on the website, add public key fingerprint to [static/keys](https://github.com/theforeman/theforeman.org/tree/gh-pages/static/keys)
- * [ ] create [releases/<%= release %>/RPM-GPG-KEY-foreman](https://yum.theforeman.org/releases/<%= release %>/RPM-GPG-KEY-foreman) on yum.theforeman.org
+* [ ] [Generate](https://github.com/ekohl/theforeman-rel-eng/blob/master/generate_gpg), [backup](https://github.com/ekohl/theforeman-rel-eng/blob/master/export_gpg_private), [sign](https://github.com/ekohl/theforeman-rel-eng/blob/master/sign_gpg) and [upload](https://github.com/ekohl/theforeman-rel-eng/blob/master/upload_gpg) a GPG key. See [GPG_Keys](https://projects.theforeman.org/projects/foreman/wiki/GPG_Keys) if needed
```
ehelms: This seems like individual steps that would be good as a set of sub-bullets.

ekohl: I've always ran them right after each other, but I like separate individual steps in case something goes wrong.

evgeni: This should live in the @theforeman org, right, right? :)

ekohl: Yes, but I'd prefer to do that as a separate step :)

PR #164, opened 2019-04-23
Updates for 3.11 and 3.12

jturel:


jturel:
@chris1984 FYI, I'm adding 3.11.1 as a sub-release of 3.12 to make sure nothing is missed while cherry-picking (ie things that need to go into 3.11.1 and thereby 3.12 *after* 3.12 branching)

PR #172, opened 2019-04-25
Fix 1.22 mash scripts

ekohl:


mash_scripts/foreman/1.22.0/foreman-plugins-1.22-rhel7-dist.mash
```diff
@@ -7,6 +7,6 @@ multilib = False
 multilib_method = devel
 tag = foreman-plugins-1.22-rhel7-dist
 inherit = True
-strict_keys = True
+strict_keys = False
```
chris1984: ACK, the only question I had was why are we turning this off?

ekohl: We've never signed plugin repositories.

chris1984: LGTM then

PR #182, opened 2019-06-24
Bump 3.12 to 3.12.1

chris1984:
@jturel does this look right?


jturel:
Instead of removing 3.11.1 as a prior_release, 3.11.1 & 3.11.2 should both be under prior_releases. Other than that, LGTM!

chris1984:
@jturel thanks for the review :) Updated!

PR #185, opened 2019-07-29
add Foreman 1.23 config

evgeni:


evgeni:
@jjeffers added more Katello 3.13 related changes.

PR #186, opened 2019-07-29
add mash scripts from Foreman 1.23 and Katello 3.13

evgeni:


mash_scripts/foreman/1.23.0/foreman-1.23-rhel7-dist.mash
```diff
@@ -0,0 +1,12 @@
...
+rpm_path = %(arch)s/os/Packages
+repodata_path = %(arch)s/os/
+source_path = source/SRPMS
+debuginfo = False
+multilib = False
+multilib_method = devel
+tag = foreman-1.23-rhel7-dist
+inherit = True
+strict_keys = True
+keys = 4674C97BD8C2A3EF63BCB20788BB5C467B5B366A
```
ekohl: Pretty sure in the past I had issues with the long key and needed the short version (6 chars) because koji doesn't do full keys. The reason I have a short one in the settings of my tool (https://github.com/ekohl/theforeman-rel-eng/blob/37b6d9d92448a85cb8acfbb5ec65cd8f31e6a30a/releases/1.23/settings#L3). Maybe this also needs to be changed in the tool_belt main config from which this was generated.

evgeni: updated.

PR #190, opened 2019-07-31
Update foreman release procedure

tbrisker:


procedures/foreman/release.md.erb
```diff
@@ -1,15 +1,15 @@
...
   * [ ] Release notes: bullet point list by category of all changes, include link to bug numbers. You can auto-generate changes using the [release notes script](https://github.com/theforeman/theforeman.org/blob/gh-pages/scripts/release_notes.rb).
   * [ ] CLI release notes are taken from the hammer-cli and hammer-cli-foreman changelogs
   * [ ] Link to installer changelogs and note versions being used
-  * [ ] Update installer options section using the [get-params script](https://github.com/theforeman/theforeman.org/blob/gh-pages/scripts/installer/get-params)
-* [ ] Get the apipie doc and place it in [api](https://github.com/theforeman/theforeman.org/tree/gh-pages/api)/<%= short_version %>
-
+<% if (is_rc || full_version.end_with?('0')) %>  * [ ] Headline features: half a dozen important features with a few sentences description each
+  * [ ] Upgrade notes: all important notices that users must be aware of before upgrading
+* [ ] Update installer options section using the [get-params script](https://github.com/theforeman/theforeman.org/blob/gh-pages/scripts/installer/get-params)
```
ekohl: I'd actually say this should always be done. Ideally it would be no difference, but otherwise you know.

tbrisker: specifically the installer options or all of them?

ekohl: I'd prefer both installer params and apipie. Ideally it should be no diff (though I think apipie isn't that stable).

procedures/foreman/release.md.erb
```diff
@@ -1,15 +1,15 @@
...
+* [ ] [Generate](https://github.com/theforeman/theforeman.org#updating-api-auto-generated-docs-by-apipie) the apipie doc and place it in the api/<%= short_version %> folder
```
ekohl:
```suggestion
* [ ] [Generate](https://github.com/theforeman/theforeman.org#updating-api-auto-generated-docs-by-apipie) the apipie doc and place it in the api/<%= short_version %> directory
```

procedures/foreman/release.md.erb
tbrisker: apipie is actually pretty stable, and shouldn't change in .z versions normally. Also generating the docs takes a couple of hours on my computer, so i'd prefer not to redo it every time. I'll update this though to remind to check that in z stream.

procedures/foreman/release.md.erb
tbrisker: fixed

procedures/foreman/release.md.erb
ekohl: In meant in the generated HTML. If you rerun the same job twice, it may change formatting I think due to changing examples.

tbrisker:
ping, updated according to comments

procedures/foreman/release.md.erb
ekohl: During 1.23 we noticed the script only works on already released bits. Probably good to include in the instructions somewhere.

tbrisker: added comment about that

PR #191, opened 2019-08-01
correct katello configs

evgeni:
they had not have the right parents set

### DIFF
--- a/configs/katello/3.12.yaml
+++ b/configs/katello/3.12.yaml
@@ -10,7 +10,7 @@
   :3.12.2:
     :redmine_version_id: 1053
   :3.12.1:
-    :redmine_version_id: 1031 
+    :redmine_version_id: 1031
   :3.12.0:
     :redmine_version_id: 993
 :code_name: Iron Stout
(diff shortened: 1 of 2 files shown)


evgeni:
@chris1984 @jjeffers it seems the tool belt configs for 3.12 and 3.13 were generated off the nightly config and not the N-1 release one, which resulted in wrong inheritance and copied packages.

compare [3.11](http://koji.katello.org/koji/taginfo?tagID=494) to [3.12](http://koji.katello.org/koji/taginfo?tagID=544) and [3.13](http://koji.katello.org/koji/taginfo?tagID=591).

@ehelms @jlsherrill @jturel if you want to review that too? :)

evgeni:
(I prefer to get explicit ACKs from Katello release owners before merging this)

chris1984:
That makes sense @evgeni thanks for catching the issue so we can go into 3.13 with the correct configs.

jturel:
Were there any consequences from this mistake? As far as I know 3.12 has been fine (not suggesting we don't fix it!).

evgeni:
> Were there any consequences from this mistake? As far as I know 3.12 has been fine (not suggesting we don't fix it!).

Well, it's complicated. If you look at the 3.12 vs 3.11 repos, 3.12 don't inherit from Foreman (but it seems not to cause any problems?) and it inherits from the "nightly" thirdparty repo, instead of the released one. There were no changes to that thirdparty repo, so it's not a huge problem right now. But it's still not nice.

So I'd say no direct consequence today, but possible in the future.

evgeni:
ftr, for 3.13, I removed the tag from koji completely and recreated it.

for 3.12 I'd just let toolbelt fix the inheritance, but not touch anything else

PR #192, opened 2019-08-02
Do not check plugin GPG keys

ekohl:


evgeni:
is there a config that needs tuning for that?

ekohl:
Not sure the mash builder is flexible enough for this. https://github.com/theforeman/tool_belt/blob/master/lib/tool_belt/mash_scripts_builder.rb looks like it has a global strict option and would need per tag overrides.

PR #194, opened 2019-08-06
Update Katello 3.13 config for correct releases

chris1984:


chris1984:
@zjhuntin can this get a merge please when you get a moment?

configs/katello/3.13.yaml
```diff
@@ -31,6 +31,9 @@
     :branch: 3.5.1
     :repo: https://github.com/Katello/katello-host-tools.git
 :ignores:
+  - 25114 # Already merged into 3.12
```
jturel: I haven't checked but if these are for the 3.12.0 or 3.12.1 releases we can just update the config to remove those two from the prior_releases in the config. They don't need to be reflected here because they are already GA. 3.12.2 should be left as a prior release since it's still in flight.

ekohl: What's the definition of a prior release? I'd expect that 3.12.0 and 3.12.1 are prior since they were released before 3.13.0 but 3.12.2 isn't because it's not released yet.

jturel: My definition based on what I've experienced in this repo is that a prior release is one that we want to track for cherry-picks that we also want to include in the release stream for the version in the config.

For example:

3.13.0 is in flight but 3.12.2 is also in flight. If there is a new change on master *today* for 3.12.2 we want it to be picked up by `cherry-pick` for inclusion in 3.13.0 so that's why we indicate it as prior. While 3.12.0 and 3.12.1 are technically 'prior' they are already released and nothing should be getting marked for cherry-picking from them.

Let me know if I'm forgetting a use case.

ekohl: Without that much actual knowledge my impression was that it verified everything that was in a prior release, is also in this release. The goal should be that if something was cherry picked into 3.12.1 is also in 3.13.0.

Ideally we should actually be setting the `Fixed in version` in Redmine. If it's scheduled for cherry pick, you set for a future release (i.e., 3.12.2). Then when you release 3.12.2, this script would very that all commits are actually present. One benefit is that you exactly know into which z-stream it was released (or will be) when you find a Redmine issue as a user.

I think this is also how @tbrisker works.

chris1984:
Sounds good, I will update the pr

configs/katello/3.13.yaml
jturel: I think we are in agreement about what the goal of prior releases are.

To your example, though: 3.12.1 was already released by the time 3.13.0 branched so by definition everything that is on 3.12.1 (and ever will be) is in 3.13.0 which is why I am suggesting to remove it and 3.12.0 from this release config.

ekohl: If it's by definition present, then why does it show up here? Or is this also for the changelog?

jturel: Good question. Maybe 'by definition' was too broad because there could be reasonable explanations for them to show up. I looked at the issues above to find:

- one of them is a docs pr which was merged. theforeman.org doesn't get branched but could potentially add it to the config:
```
  :theforeman.org:                                                                                                                                                           
    :branch: master                                                                                                                                                          
    :repo: https://github.com/theforeman/theforeman.org
```

- another issue had a closed packaging PR in addition to the one that got merged. I removed that PR from the issue and it not longer shows up

- last one (25114) has a PR in Katello and one in Bastion and i think that may be confusing the cherry-picker perhaps because bastion isn't in the config (and shouldn't be since it's in Katello proper now)

That said, 3.12.2 should be the only prior release in the 3.13 config and #25114 should be checked by @chris1984  for presence in 3.12 code It might be obsoleted because of the move of bastion into Katello.

tbrisker: > I think this is also how @tbrisker works.

Actually the process for Foreman is a bit different than the Katello one - we don't do all of the cherry-picks at once right before the release, but rather one by one (either me manually doing it or by PRs against the stable branch) during the full cycle. Once a commit has been cherry-picked to the stable branch, we set the "fixed in" field for the next release in that stream. 
We use "target version" to indicate fixes that we want to get into a certain release, prior to releasing I make sure there aren't any changes that were targeted for the released but not fixed in it - which is an easy query in RM, and than either block the release for them, cherry-pick the fix, or decide to push it to the next release.

chris1984:
@jturel updated

chris1984:
@ekohl can this get a merge please?

PR #197, opened 2019-08-12
make katello 3.13 use foreman 1.23 signing key

evgeni:


evgeni:
@jjeffers @jlsherrill this is the corresponding tool belt PR to sync up configs for Katello 3.13

PR #198, opened 2019-08-13
Add Katello 3.12.3 release to toolbelt

chris1984:
@jturel here is the pr for 3.12.3

### DIFF
--- a/configs/katello/3.12.yaml
+++ b/configs/katello/3.12.yaml
@@ -7,6 +7,8 @@
   :3.11.1:
     :redmine_version_id: 992
 :releases:
+  :3.12.3:
+    :redmine_version_id: 1082
   :3.12.2:
     :redmine_version_id: 1053
   :3.12.1:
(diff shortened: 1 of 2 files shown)


configs/katello/3.12.yaml
```diff
@@ -4,9 +4,9 @@
 :prior_releases:
   :3.11.2:
     :redmine_version_id: 1038
-  :3.11.1:
```
jturel: I don't think there's a reason to remove 3.11.1 from priors

configs/katello/3.13.yaml
```diff
@@ -2,8 +2,8 @@
 :project: katello
 :github_org: katello
 :prior_releases:
-  :3.12.2:
-    :redmine_version_id: 1053
+  :3.12.3:
```
jturel: 3.12.3 should be an additional item rather than a replacement for 3.12.2

chris1984:
@jturel updated

chris1984:
@ekohl can this get a merge when you get a free moment please?

jlsherrill:
Thanks @chris1984 !

PR #209, opened 2019-10-28
add Foreman 1.24 config

zjhuntin:


pcreech:
The config looks good to me, bu would like an additional ack from someone else on that part.

Also, maybe we should also add the mash_scripts/1.24.0/*.mash files in this PR?

zjhuntin:
@pcreech added

mash_scripts/foreman/1.24.0/foreman-client-1.24-fedora28.mash
```diff
@@ -0,0 +1,12 @@
+[foreman-client-1.24-fedora28]
```
ekohl: F28 is EOL. Do we really want to build packages for this? I think we also dropped it in nightly configs some time ago.

zjhuntin: ACK, i'll remove it

configs/foreman/1.24.yaml
```diff
@@ -0,0 +1,536 @@
...
+  :1.24.0:
+    :redmine_version_id: 1070
+:mash_scripts:
+  - foreman-1.24-rhel7-dist
+  - foreman-plugins-1.24-rhel7-dist
+  - foreman-rails-1.24-rhel7
+  - foreman-client-1.24-rhel7
+  - foreman-client-1.24-rhel6
+  - foreman-client-1.24-rhel5
+  - foreman-client-1.24-fedora28
```
ekohl: This should be dropped

configs/foreman/1.24.yaml
```diff
@@ -0,0 +1,536 @@
...
+  - foreman-client-1.24-fedora29
+  - foreman-client-1.24-sles11
+  - foreman-client-1.24-sles12
+:repos: []
+:gpg_key: 7B5B366A
```
ekohl: This is the 1.23 GPG key

PR #212, opened 2019-10-30
Update RPM packaging instructions

ekohl:


procedures/foreman/branch.md.erb
```diff
@@ -79,8 +79,8 @@
   * [ ] [foreman-installer](https://github.com/theforeman/foreman-installer)
   * [ ] [foreman-selinux](https://github.com/theforeman/foreman-selinux)
   * [ ] [smart-proxy](https://github.com/theforeman/smart-proxy)
-* [ ] Update foreman-packaging:
-  * [ ] rpm/develop: Update `rel-eng/{releasers.conf,tito.props}`, `packages/foreman/foreman{,-{installer,proxy,release,selinux}}/*.spec`
+* [ ] Update foreman-packaging to <%= develop %>:
+  * [ ] rpm/develop: `sed -i 's/fm<%= release.tr('.', '_') %>/fm<%= develop.tr('.', '_') %>/g' rel-eng/{releasers.conf,tito.props} && obal update --version <%= develop %>.0 --release 1 --prerelease develop --changelog '- Bump version to <%= develop %>-develop' foreman{,-{installer,proxy,release,selinux}}`
```
ekohl: Depends on https://github.com/theforeman/obal/pull/215

ekohl: Looks like this fails on annexing the source which it shouldn't attempt to do.

ekohl: Yes. Wonder whether we should have a `--spec-only` parameter or a separate subcommand.

ekohl:
Should shouldn't have been merged. Due to the annex the command actually doesn't work and fails after updating one spec.

PR #213, opened 2019-11-02
Fix RPM instructions

ekohl:
0a955d04273dc2d97b5c8b53ff0fa158e1264cb1 changed this, but it doesn't work because obal can't handle packages without sources. This causes the fetch to break and the command quits after the first package.


evgeni:
What are "packages without sources"? Like foreman that has no sources in git annex for develop?

ekohl:
Correct. The alternative would be to modify obal to skip managing the source files and just update the spec. I'm fine either way, but just don't want broken instructions.

PR #214, opened 2019-11-04
katello 3.14 config

evgeni:


configs/foreman/1.24.yaml
```diff
@@ -15,7 +15,7 @@
   - foreman-client-1.24-sles11
   - foreman-client-1.24-sles12
 :repos: []
-:gpg_key: 7B5B366A
+:gpg_key: 5AA9BCAD
```
evgeni: @zjhuntin this seems to be an oversight in the 1.24 config?

evgeni:
```
% koji untag-build katello-3.14-rhel7 tfm-rubygem-hammer_cli_katello-0.20-0.1.pre.master.20191025142410gite38004e.el7 katello-3.15.0-0.1.master.el7 katello-repos-3.15.0-0.1.nightly.el7 tfm-rubygem-katello-3.15.0-0.1.pre.master.20191104142047git9aa9fa8.el7
```

to clean up builds that shouldn't be there

configs/foreman/1.24.yaml
zjhuntin: It is, Something I needed to change, but forgot.

PR #215, opened 2019-11-04
Updates config for katello 3.13.1 and 3.13.2.

jjeffers:
This is the config change for 3.13.1, which apparently I didn't push earlier.


jturel:
Go ahead and add 3.13.2 also

PR #216, opened 2019-11-05
fixup katello 3.14 config

evgeni:


evgeni:
these changes are really by @jturel, so gonna merge them as is :)

PR #223, opened 2019-11-20
Katello pulpcore mash config

pcreech:


evgeni:
where is the config that generated that? :)

evgeni:
or is that straight from https://github.com/theforeman/tool_belt/blob/master/configs/katello/nightly.yaml unmodified?

pcreech:
.. it was hand created :D ....  Missed the fact that we had generators for it

pcreech:
@evgeni updated

evgeni:
the generated file looks like what I would expect to to look like

PR #228, opened 2019-12-12
add pulpcore update procedure

evgeni:


procedures/pulpcore/update.md.erb
```diff
@@ -0,0 +1,17 @@
...
+pip3 install pulpcore pulp-file pulp-container
+pip3 freeze |sed '/certifi/d'
+```
+* [ ] Save the `pip` output to a file
+
+# Update `foreman-packaging`
+
+* [ ] Use the list of Python packages and update their RPM packages:
+```bash
+while read line; do pkg=$(echo ${line} |cut -f1 -d=); version=$(echo ${line} | cut -f3 -d=); ./add_pypi_package.sh ${pkg} ${version} katello-pulpcore-nightly-el7 pulpcore; done < THE_PIP_LIST_FILE
```
evgeni: this needs https://github.com/theforeman/foreman-packaging/pull/4478 or some other solution to not regenerate existing packages

procedures/pulpcore/update.md.erb
```diff
@@ -0,0 +1,17 @@
+# Collect list of Python packages
+
+* [ ] Execute the following script in an empty EL7 container:
+```bash
```
evgeni: this feels like it should be stored as a script somewhere, but I couldn't think of a good place?

procedures/pulpcore/update.md.erb
ekohl: Using bash instead of cut:
```suggestion
while read line; do pkg=${line%==*} ; version=${line#*==} ; ./add_pypi_package.sh ${pkg} ${version} katello-pulpcore-nightly-el7 pulpcore; done < THE_PIP_LIST_FILE
```

The very short version:

```suggestion
while read line; do ./add_pypi_package.sh ${line%==*} ${line#*==} katello-pulpcore-nightly-el7 pulpcore; done < THE_PIP_LIST_FILE
```

procedures/pulpcore/update.md.erb
ekohl: IMHO foreman-packaging is still a good place.

procedures/pulpcore/update.md.erb
evgeni: Yeah, the long version is good :)

procedures/pulpcore/update.md.erb
evgeni: `collect_pulpcore_packages.sh` in the root of the repo?

ekohl: Feels like we may need a `script` directory

evgeni: Yes it does, but then again, I don't actually want to branch those scripts…

ekohl: We already do that with many now and they will fail after branching. By adding one more you're not making it worse and at least keep all packaging actions in two places (f-packaging and obal). If obal knew how to deal with adding packages, that would be the correct place.

ehelms: Could this whole procedure be reduced down to a combination of an Ansible playbook and a container? Perhaps an obal command?

evgeni: something like https://github.com/evgeni/pulpcore-packaging/blob/master/Dockerfile + https://github.com/evgeni/pulpcore-packaging/blob/master/_generate_deps.sh and then running with `podman run --rm -v $(pwd)/foreman-packaging/:/app/foreman-packaging/:Z <container>`?

ekohl: I'd probably simplify it to (still using podman):
```Dockerfile
FROM centos:7
RUN yum install -y postgresql-devel gcc python3 python3-devel python3-pip
RUN pip3 install pulpcore pulp-file pulp-container
```

```bash
#!/bin/bash -e

PULPCORE_REQUIREMENTS="pulpcore-requirements.txt"
CONTAINER_TAG=pulpcore-packaging

podman build --pull-always --tag $CONTAINER_TAG .
podman run --rm $CONTAINER_TAG pip3 freeze | sed '/certifi/d' > $PULPCORE_REQUIREMENTS
podman rmi $CONTAINER_TAG
```

Alternative using buildah:
```bash
#!/bin/bash -e

PULPCORE_REQUIREMENTS="pulpcore-requirements.txt"

newcontainer=$(buildah from centos:7)
buildah run $newcontainer yum install -y postgresql-devel gcc python3 python3-devel python3-pip
buildah run $newcontainer pip3 install pulpcore pulp-file pulp-container
buildah run $newcontainer pip3 freeze | sed '/certifi/d' > $PULPCORE_REQUIREMENTS
buildah rm $newcontainer
```

evgeni: this would simplify it by dropping the `add_pypi_package` step ;-)

evgeni: I've updated https://github.com/evgeni/pulpcore-packaging with a better dockerfile and a readme :)

ehelms: So this procedure should get updated to instruct using that repository?

evgeni: That's what I did now. Yeah.

PR #229, opened 2020-01-08
EL8 client koji config

pcreech:


evgeni:
Looks good to me. I think the only thing standing out is the naming with *el8* instead of *rhel8*, which I am fine with but wanted to hear another opinion. @ekohl @ehelms?

pcreech:
FWIW, this is following a convention we started with pulpcore's tag, switching to an "elx" instead of a "rhelx" convention

`katello-pulpcore-nightly-el7
katello-pulpcore-nightly-el7-build
`

evgeni:
Yeah, that's one of the reasons I'm completely fine with it. Just with Katello it was completely new target, here it's just adding an arch to an existing target

ekohl:
I like el7/el8 because it's also what we use for https://yum.theforeman.org/ directories. It makes syncing easier if we don't need to transform/map.

pcreech:
> Yeah, that's one of the reasons I'm completely fine with it. Just with Katello it was completely new target, here it's just adding an arch to an existing target

Yup, just recording the reasoning on the PR

PR #230, opened 2020-01-13
For katello-3.13.4 release.

jjeffers:
Apparently this config change was left out. As we're producing a 3.13.4 we'll need it.


jturel:
Needs the 3.13.4 release  + the ignore we talked about :)

PR #232, opened 2020-01-14
Skip features heading if empty

jjeffers:
This PR introduces a check for empty feature sets in a changelog command. If so, then the featueres heading is skipped.

### DIFF
--- a/test/format_entries_test.rb
+++ b/test/format_entries_test.rb
@@ -1,0 +1,49 @@
+require 'minitest/autorun'
+require 'minitest/mock'
+require 'tool_belt'
+require 'mocha/minitest'
+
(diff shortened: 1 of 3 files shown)


evgeni:
Looks good, but needs a rebase :)

test/format_entries_test.rb
```diff
@@ -0,0 +1,49 @@
+require 'minitest/autorun'
+require 'minitest/mock'
+require 'tool_belt'
+require 'mocha/minitest'
+
+class FormatEntriesTest < MiniTest::Test
```
jturel: Love seeing a test. What do you think about changing this from `FormatEntriesTest` to `ChangelogTest` to broaden the scope a bit, and keep future tests for the class in one file?

PR #236, opened 2020-02-05
Katello el8 koji config

pcreech:


configs/katello/nightly.yaml
```diff
@@ -35,6 +35,27 @@
       - centos-sclo-rh-rhel-7
       - centos-7-server-updates
       - centos-7-server
+  - name: katello-nightly-el8
+    based_off: null
+    helper_tags:
+      katello-thirdparty-el8: null
+      katello-nightly-el8-override: null
+    build_target: katello-nightly-el8-build
+    build_package_group_source_tag: katello-nightly-el7-build
```
evgeni: nope
```suggestion
    build_package_group_source_tag: katello-nightly-el8-build
```

pcreech: good catch.  Updated

configs/katello/nightly.yaml
```diff
@@ -78,3 +99,47 @@
...
         - shadow-utils
+  - name: katello-pulpcore-nightly-el8
+    based_off: null
+    arches:
+     - x86_64
+    build_target: katello-pulpcore-nightly-el8-build
+    external_repos:
+      - centos8-baseos
+      - centos8-appstream
+    build_groups:
```
ehelms: Do we need to specify this for the katello build tag above?

pcreech: I don't think so.  I followed the pattern with the prior katello el7 build tag, and it doesn't seem that it was there.

This could be due to an environment not entirely fleshed out as well.  I'll double check today.

ehelms: I think we never did that because it was already built out and for releases we often 'copied'. But for fresh tags, I find it worthwhile to specify the build tag components exactly as you've done here.

pcreech: So, after digging into it, and looking at katello-nightly-rhel7 on koji... it looks like the build_groups were set not by toolbelt, but manually somehow.  (it's not inheriting it's build groups)

Will add it to el8

PR #237, opened 2020-02-07
Add deprecation warning scan

johnpmitsch:
There is a related discussion here:
https://community.theforeman.org/t/deprecation-warnings-how-to-handle-better/17233/9

We will often forget to remove deprecation warnings in the release we say we will. This commit adds the parser gem and a class inheriting from ::Parser::AST::Processor which will scan for deprecation warning methods and any versions in the arguments. It will compare them with the passed in version and print any outdated deprecation locations.

### DIFF
--- a/lib/tool_belt/commands/check_deprecation_warnings.rb
+++ b/lib/tool_belt/commands/check_deprecation_warnings.rb
@@ -1,0 +1,64 @@
+module ToolBelt
+  module Command
+    class CheckDeprecationWarningsCommand < Clamp::Command
+      parameter 'repo_path', 'Path (relative to tool_belt) of repository to check'
+      option '--version', 'VERSION', 'Release version to check for in x.y format', required: true
+      option '--paths', 'PATHS',
+             'Patterns within the repository to search, separated by a comma.',
+             default: 'app/**/*.rb,lib/**/*.rb'
+
+      def execute
+        root_path = File.expand_path("../../../../#{repo_path}", __FILE__)
+        globs_to_scan = paths.split(',').map { |path| File.expand_path(path, root_path) }
+        puts "\nScanning for deprecation warnings in #{globs_to_scan.join(', ')}\n\n"
+        paths_to_scan = Dir.glob(globs_to_scan)
+
+        finder = ToolBelt::OutdatedDeprecationWarningFinder.new(paths_to_scan, version)
+        finder.scan_for_deprecations
+        current_outdated = finder.deprecations.select { |dep| dep[:outdated] == true }
+        next_outdated = finder.deprecations.select { |dep| dep[:outdated_next_version] == true }
+
+        if current_outdated.any?
+          puts current_outdated_found
+          current_outdated.each { |outdated_dep| puts current_outdated_message(outdated_dep, finder.current_version) }
+        else
+          puts "No currently outdated deprecation warnings found!\n".green
+        end
+
+        if next_outdated.any?
+          puts next_outdated_found(finder.next_minor_version)
+          next_outdated.each { |outdated_dep| puts next_outdated_message(outdated_dep, finder.next_minor_version) }
+        else
+          puts "No deprecation warnings found that will be outdated in the next version!\n".green
+        end
+      end
+
+      private
+
+      def current_outdated_found
+        'Deprecation warnings were found with versions that are outdated, please check below and ' \
+        'ensure both the deprecated functionality and the deprecation warning are removed if ' \
+        "the deprecation is outdated in this release\n".red
+      end
+
+      def next_outdated_found(version)
+        "Deprecation warnings were found marked for removal for the *next* release version #{version}. " \
+        "Please check below and file issues aligned to #{version} to remove the functionality and the " \
+        'deprecation warning. This will ensure the functionality is removed in time for the next ' \
+        "release #{version}. The deprecation warning should stay in this release, only file issues for " \
+        "its removal at this time.\n".yellow
+      end
+
+      def current_outdated_message(outdated_dep, version)
+        "Deprecation warning at #{outdated_dep[:file]} line #{outdated_dep[:line]} is marked for " \
+        "removal in version #{outdated_dep[:version]}, which is less than or equal to the specified " \
+        "version of #{version}.\n\n"
+      end
+
+      def next_outdated_message(outdated_dep, _version)
+        "Deprecation warning at #{outdated_dep[:file]} line #{outdated_dep[:line]} is marked for " \
+        "removal in version #{outdated_dep[:version]}, which is the *next* release version.\n\n"
+      end
+    end
+  end
+end
--- a/lib/tool_belt/commands/check_deprecation_warnings.rb
+++ b/lib/tool_belt/commands/check_deprecation_warnings.rb
@@ -1,20 +1,22 @@
 module ToolBelt
   module Command
     class CheckDeprecationWarningsCommand < Clamp::Command
-      parameter 'repo_path', 'Path (relative to tool_belt) of repository to check'
-      option '--version', 'VERSION', 'Release version to check for in x.y format', required: true
+      parameter 'config_file', 'Release configuration file'
       option '--paths', 'PATHS',
              'Patterns within the repository to search, separated by a comma.',
              default: 'app/**/*.rb,lib/**/*.rb'
 
       def execute
-        root_path = File.expand_path("../../../../#{repo_path}", __FILE__)
-        globs_to_scan = paths.split(',').map { |path| File.expand_path(path, root_path) }
-        puts "\nScanning for deprecation warnings in #{globs_to_scan.join(', ')}\n\n"
+        config = ToolBelt::Config.new(config_file)
+        options = config.options
+        repo_path = "repos/#{options.namespace}/#{options.project}/"
+        raise "Repo path #{repo_path} not found!" unless File.exist?(repo_path)
+
+        globs_to_scan = paths.split(',').map { |path| File.expand_path(path, repo_path) }
         paths_to_scan = Dir.glob(globs_to_scan)
 
-        finder = ToolBelt::OutdatedDeprecationWarningFinder.new(paths_to_scan, version)
-        finder.scan_for_deprecations
+        puts "\nScanning for deprecation warnings in #{globs_to_scan.join(', ')}\n\n"
+        finder = ToolBelt::OutdatedDeprecationWarningFinder.new(paths_to_scan, options.release)
         current_outdated = finder.deprecations.select { |dep| dep[:outdated] == true }
         next_outdated = finder.deprecations.select { |dep| dep[:outdated_next_version] == true }
 
@@ -27,7 +29,7 @@ module ToolBelt
 
         if next_outdated.any?
           puts next_outdated_found(finder.next_minor_version)
-          next_outdated.each { |outdated_dep| puts next_outdated_message(outdated_dep, finder.next_minor_version) }
+          next_outdated.each { |outdated_dep| puts next_outdated_message(outdated_dep) }
         else
           puts "No deprecation warnings found that will be outdated in the next version!\n".green
         end
@@ -55,7 +57,7 @@ module ToolBelt
         "version of #{version}.\n\n"
       end
 
-      def next_outdated_message(outdated_dep, _version)
+      def next_outdated_message(outdated_dep)
         "Deprecation warning at #{outdated_dep[:file]} line #{outdated_dep[:line]} is marked for " \
         "removal in version #{outdated_dep[:version]}, which is the *next* release version.\n\n"
       end
--- a/lib/tool_belt/deprecation_warning_parser.rb
+++ b/lib/tool_belt/deprecation_warning_parser.rb
@@ -2,7 +2,7 @@ require 'parser/current'
 
 module ToolBelt
   class DeprecationWarningParser < ::Parser::AST::Processor
-    attr_accessor :versions_found
+    attr_reader :versions_found
 
     def initialize(file)
       super()
--- a/lib/tool_belt/deprecation_warning_parser.rb
+++ b/lib/tool_belt/deprecation_warning_parser.rb
@@ -1,0 +1,38 @@
+require 'parser/current'
+
+module ToolBelt
+  class DeprecationWarningParser < ::Parser::AST::Processor
+    attr_accessor :versions_found
+
+    def initialize(file)
+      super()
+      @file = file
+      @versions_found = []
+    end
+
+    def on_begin(node)
+      node.children.each { |c| process(c) }
+    end
+
+    def on_send(node)
+      method_class, method_name, *args = node.children
+      version_regex = /\d+\.\d+/
+
+      if %i[api_deprecation_warning deprecation_warning].include?(method_name)
+        *method_classes = method_class.children if method_class&.children
+        if method_classes.include?(:Deprecation)
+          args.select! { |arg| arg.to_s =~ version_regex }
+          if args.any?
+            found_version = args.first.children.first
+            extracted_version = found_version.scan(version_regex).first
+            @versions_found << {
+              version: extracted_version,
+              file: @file,
+              line: node.loc.line
+            }
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/tool_belt/outdated_deprecation_warning_finder.rb
+++ b/lib/tool_belt/outdated_deprecation_warning_finder.rb
@@ -1,0 +1,61 @@
+module ToolBelt
+  class OutdatedDeprecationWarningFinder
+    attr_accessor :deprecations
+    attr_accessor :current_version
+    attr_accessor :next_minor_version
+
+    def initialize(paths, version)
+      @paths = paths
+      @version = version
+      @deprecations = []
+      error_if_empty(paths)
+    end
+
+    def scan_for_deprecations
+      outdated_deprecations = []
+
+      @paths.select do |file|
+        ast = ::Parser::CurrentRuby.parse(File.read(file))
+        processor = ToolBelt::DeprecationWarningParser.new(file)
+        processor.process(ast)
+
+        next unless processor.versions_found.any?
+
+        processor.versions_found.each do |version_found|
+          if Gem::Version.new(version_found[:version]) <= current_version
+            version_found[:outdated] = true
+            @deprecations << version_found
+          elsif Gem::Version.new(version_found[:version]) == next_minor_version
+            version_found[:outdated_next_version] = true
+            @deprecations << version_found
+          end
+        end
+      end
+    end
+
+    def next_minor_version
+      bump_minor_version(current_version)
+    end
+
+    def current_version
+      Gem::Version.new(@version)
+    end
+
+    private
+
+    def error_if_empty(paths_to_scan)
+      if paths_to_scan.empty?
+        raise "Couldn't find any files! Please check the path and try again"
+      end
+    rescue RuntimeError => e
+      puts e.message.red
+      exit 1
+    end
+
+    def bump_minor_version(version)
+      version_segments = version.segments
+      version_segments[1] += 1
+      Gem::Version.new(version_segments.join('.'))
+    end
+  end
+end
--- a/lib/tool_belt/outdated_deprecation_warning_finder.rb
+++ b/lib/tool_belt/outdated_deprecation_warning_finder.rb
@@ -1,19 +1,16 @@
 module ToolBelt
   class OutdatedDeprecationWarningFinder
     attr_accessor :deprecations
-    attr_accessor :current_version
-    attr_accessor :next_minor_version
 
     def initialize(paths, version)
       @paths = paths
       @version = version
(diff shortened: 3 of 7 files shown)


johnpmitsch:
Some examples of usage:

Katello:
```
[jomitsch@jomitsch tool_belt{dep_warm}]$ ./tools.rb check-deprecation-warnings --version 3.14 repos/katello/3.14.1/katello

Scanning for deprecation warnings in /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/**/*.rb

Deprecation warnings were found with versions that are outdated, please check below and ensure both the deprecated functionality and the deprecation warning are removed if the deprecation is outdated in this release

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/host_subscriptions_controller.rb line 161 is marked for removal in version 3.5, which is less than or equal to the specified version of 3.14.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/repositories_controller.rb line 409 is marked for removal in version 3.3, which is less than or equal to the specified version of 3.14.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/content_view_versions_controller.rb line 291 is marked for removal in version 3.4, which is less than or equal to the specified version of 3.14.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/subscriptions_controller.rb line 204 is marked for removal in version 2.6, which is less than or equal to the specified version of 3.14.
```
Foreman (checking 1.24, no outdated dep warnings show)
```
[jomitsch@jomitsch tool_belt{dep_warm}]$ ./tools.rb check-deprecation-warnings --version 1.24 repos/foreman/1.24.0/foreman/

Scanning for deprecation warnings in /home/jomitsch/tool_belt/repos/foreman/1.24.0/foreman/app/**/*.rb

No outdated deprecation warnings found!
```

Foreman (checking a future version to show outdated dep warnings)
```
[jomitsch@jomitsch tool_belt{dep_warm}]$ ./tools.rb check-deprecation-warnings --version 2.0 repos/foreman/1.24.0/foreman/

Scanning for deprecation warnings in /home/jomitsch/tool_belt/repos/foreman/1.24.0/foreman/app/**/*.rb

Deprecation warnings were found with versions that are outdated, please check below and ensure both the deprecated functionality and the deprecation warning are removed if the deprecation is outdated in this release

Deprecation warning at /home/jomitsch/tool_belt/repos/foreman/1.24.0/foreman/app/helpers/application_helper.rb line 185 is marked for removal in version 1.27, which is less than or equal to the specified version of 2.0.

Deprecation warning at /home/jomitsch/tool_belt/repos/foreman/1.24.0/foreman/app/models/concerns/orchestration/dhcp.rb line 96 is marked for removal in version 1.25, which is less than or equal to the specified version of 2.0.
```

Error handling
```
[jomitsch@jomitsch tool_belt{dep_warm}]$ ./tools.rb check-deprecation-warnings --version 3.14 repos/katello/3.14.1/

Scanning for deprecation warnings in /home/jomitsch/tool_belt/repos/katello/3.14.1/app/**/*.rb

Couldn't find any files! Please check the path and try again
```

johnpmitsch:
We may be able to parse the repo location and version from the configuration rather than pass it in? Not sure if its possible or one way is preferred over the other, looking for some feedback :)

johnpmitsch:
I have to add the checklist entries as well, would like some guidance from those who know the process better as to where this step would fit best.

jturel:
The tool works exactly how I would expect it to - nice! Now I'm thinking about the workflow:

We're about to branch 3.15, so let's say at branch time is when we check the deprecations for 3.15.0. The drawback is that we've gone the entire 3.15 development phase with the deprecations present in the code. It'd be better to remove the deprecations & their related functionality at the very start of the dev phase so we hopefully get more testing around their removal. 

TLDR: I think the documented procedure would be to check the deprecations for the *next* release version and file issues to be addressed early in the cycle.

jturel:
cc @jjeffers you might have some input here since you're the 3.15 release owner :)

jjeffers:
If we can add this deprecation check as an automated build step for code changes, then why just do that all the time? That way, we don't have to wait until the end of the development cycle.

johnpmitsch:
> we can add this deprecation check as an automated build step for code changes,

That was the original thought, but it would have to live with the Foreman application code so it could be shared with plugins and add a new gem to Foreman (even if its just development). It started to make more sense to add here.

> That way, we don't have to wait until the end of the development cycle.

I think that is what @jturel is getting at, for example for 3.15, you would check for deprecation warnings that say functionality will be removed in 3.16, which means you would file issues to remove the functionality with the target release of 3.16. So nothing would block 3.15 and the release would be added. The timeline being (using 3.14-3.16 as an example):

1. post-3.14 branching: add deprecation warning in Katello master that X will be removed in 3.16
2. 3.15 branches and includes this deprecation warning
3. 3.15 release manager searches for N+1 deprecations, sees a deprecation warning for 3.16, and files issue to remove X and it's associated deprecation in 3.16
4. 3.15 releases with deprecation warning that X will be removed in 3.16
5. The issue is followed up on and X functionality is removed along with it's associated deprecation (this can even happen before 3.15 releases, as long as 3.15 has been branched and the redmine issue to remove X is correctly aligned to 3.16)
6. 3.16 is branched with X removed
7. 3.16 is released with X removed.

However, since we haven't been checking the deprecation warnings, we currently have a bunch of old deprecation warnings to clean up. So I don't think we would put that burden on the 3.15 release, but maybe can file some issues and try to remove them all before 3.16. Then can continue this N+1 search process

Hopefully this makes sense! In practice I think it will be simple.

I think we could have output like this to cover all bases:

```
[jomitsch@jomitsch tool_belt{dep_warm}]$ ./tools.rb check-deprecation-warnings --version 3.15 repos/katello/3.15.0/katello

Scanning for deprecation warnings in /home/jomitsch/tool_belt/repos/katello/3.15.0/katello/app/**/*.rb

Deprecation warnings were found marked for removal for the *next* release version 3.16. Please check below and file issues aligned to 3.16 to remove the functionality and the deprecation warning. This will ensure the functionality is removed in time for the next release 3.16. The deprecation warning will stay in this release, only file issues for it's removal at this time.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.15.0/katello/app/controllers/katello/api/v2/example_controller.rb line 161 is marked for removal in version 3.16, which is release following the specified version of 3.15.

Deprecation warnings were found with versions that are outdated for 3.15, please check below and ensure both the deprecated functionality and the deprecation warning are removed if the deprecation is outdated in this release

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.15.0/katello/app/controllers/katello/api/v2/host_subscriptions_controller.rb line 161 is marked for removal in version 3.5, which is less than or equal to the specified version of 3.14.
```

Thoughts?

jjeffers:
To me this seems entirely reasonable. Every action is optional and imposes no additional burden on anything. I like it.

lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -0,0 +1,68 @@
...
+
+module ToolBelt
+  module Command
+    class CheckDeprecationWarningsCommand < Clamp::Command
+
+      parameter "repo_path", "Path (relative to tool_belt) of repository to check"
+      option "--version", "VERSION", "Release version to check for", required: true
+      option "--paths", "PATHS", 
+             "Patterns within the repository to search, separated by a comma.",
+             default: "app/**/*.rb"
```
jturel: Deprecation warnings could presumably be outside of app. How about checking all .rb files?

lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def execute
+        root_path = File.expand_path("../../../../#{repo_path}", __FILE__)
+        globs_to_scan = paths.split(",").map { |path| File.expand_path(path, root_path) }
+        puts "\nScanning for deprecation warnings in #{globs_to_scan.join(", ")}\n\n"
```
jturel: I'm not sure there's a scenario where I would be checking multiple directories for the same version so maybe a single path is enough. Can you think of one?

lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -0,0 +1,68 @@
+require 'parser/current'
+require_relative '../deprecation_warning_parser.rb'
+
+module ToolBelt
+  module Command
+    class CheckDeprecationWarningsCommand < Clamp::Command
```
jturel: I'd like for this command class to be dumber than it is now: basically just taking the version and path inputs and passing them to another class for processing. That's more or less what the other commands do in this repo.

ie `scan_for_deprecations` should be somewhere else and might get called like this:
```
<snip>

outdated_deprecations = paths_to_scan.flat_map |file|
  finder = ToolBelt::DeprecationFinder.new(file)
  finder.for_version(version).select { |dep| dep[:outdated] == true }
end

# do the reporting,
```

I think messaging result text can stay in the command  since it's expected to give those messages. Let me know what you think

lib/tool_belt/commands/check_deprecation_warnings.rb
ezr-ondrej: I'd say it's ok to do only `{app,lib}/**`

johnpmitsch: I tried that and did see some utf-8 errors when checking all of the files, I didn't look too closely but can dig in a bit more as to why that was happening. 

However, I think having a select subset will keep the unnecessary errors to a minimum. App + lib (I'll add lib to the default) should be enough to cover us as that is the where majority if not all of our business logic lives. The release owner can pass in more if they would like or have a plugin with a certain structure. If there are more folders you think we should include by default, we can add them.

johnpmitsch:
> we are missing javascript deprecations here. That could rise some issues, if we start to rely on this automation.

@ezr-ondrej I didn't realize we had any, but now I see them in Foreman for the React components and certain methods provided by Foreman (through the tfm window object?), I assume these are the ones you mean? 

It would be nice to do some actual parsing and version comparing for them like in this PR. Since its not as user-facing as the API and not something plugins would have to check iiuc, how about we add a step to the foreman release process to check for them manually for now?

ekohl:
In the Foreman branching procedure we have a line that states deprecations should be removed:
https://github.com/theforeman/tool_belt/blob/308b10db42d0d45ceb98a4fbf7bc72f1bee69038/procedures/foreman/branch.md.erb#L101

IMHO a process is more important than the exact tool.

jjeffers:
> In the Foreman branching procedure we have a line that states deprecations should be removed:
 @ekohl I'm having difficulty understanding your objection here.

ekohl:
Not an objection to adding this, but rather that I think this is incomplete. By adding the invocation to the katello branching procedure, you have a much higher chance it'll actually happen. IMHO that has been the root cause of the problem. This tool just makes it easier to execute.

johnpmitsch:
I updated to refactor much of the logic to a separate class and I added a step to the katello release process, 

> Not an objection to adding this, but rather that I think this is incomplete. By adding the invocation to the katello branching procedure, you have a much higher chance it'll actually happen. IMHO that has been the root cause of the problem. This tool just makes it easier to execute.

@ekohl I had[ asked for some feedback](https://github.com/theforeman/tool_belt/pull/237#issuecomment-583485669) as to where it should go, I was waiting on that before adding the step. But I took a guess as to where it goes for Katello and you all can point me in the right direction if that is not the best place. :wink: 

Since I'm not entirely sure how Foreman wants to incorporate the tool in it's release process and there are some concerns about missing JS deprecations, I'll leave those procedures alone in this PR and let someone else update them if it's desired. I don't want to make any assumptions as I'm not very familiar with the process.

Here is the new output checking for both the current and next version in all of it's color glory:
check_deps1

lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -0,0 +1,64 @@
...
+             'Patterns within the repository to search, separated by a comma.',
+             default: 'app/**/*.rb,lib/**/*.rb'
+
+      def execute
+        root_path = File.expand_path("../../../../#{repo_path}", __FILE__)
+        globs_to_scan = paths.split(',').map { |path| File.expand_path(path, root_path) }
+        puts "\nScanning for deprecation warnings in #{globs_to_scan.join(', ')}\n\n"
+        paths_to_scan = Dir.glob(globs_to_scan)
+
+        finder = ToolBelt::OutdatedDeprecationWarningFinder.new(paths_to_scan, version)
```
jturel: nit: is 'Outdated' really clarifying the purpose of the class? I'm thinking the name is a tad long (sorry, I said 'nit' :laughing: : )

lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -0,0 +1,64 @@
...
+        "its removal at this time.\n".yellow
+      end
+
+      def current_outdated_message(outdated_dep, version)
+        "Deprecation warning at #{outdated_dep[:file]} line #{outdated_dep[:line]} is marked for " \
+        "removal in version #{outdated_dep[:version]}, which is less than or equal to the specified " \
+        "version of #{version}.\n\n"
+      end
+
+      def next_outdated_message(outdated_dep, _version)
```
jturel: unused _version arg

lib/tool_belt/deprecation_warning_parser.rb
```diff
@@ -0,0 +1,38 @@
+require 'parser/current'
+
+module ToolBelt
+  class DeprecationWarningParser < ::Parser::AST::Processor
+    attr_accessor :versions_found
```
jturel: I think it should be attr_reader?

lib/tool_belt/outdated_deprecation_warning_finder.rb
```diff
@@ -0,0 +1,61 @@
+module ToolBelt
+  class OutdatedDeprecationWarningFinder
+    attr_accessor :deprecations
+    attr_accessor :current_version
```
jturel: I see instance methods for current_version and next_minor_version so these may not be necessary

lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -0,0 +1,64 @@
...
+             default: 'app/**/*.rb,lib/**/*.rb'
+
+      def execute
+        root_path = File.expand_path("../../../../#{repo_path}", __FILE__)
+        globs_to_scan = paths.split(',').map { |path| File.expand_path(path, root_path) }
+        puts "\nScanning for deprecation warnings in #{globs_to_scan.join(', ')}\n\n"
+        paths_to_scan = Dir.glob(globs_to_scan)
+
+        finder = ToolBelt::OutdatedDeprecationWarningFinder.new(paths_to_scan, version)
+        finder.scan_for_deprecations
```
jturel: I think scan_for_deprecations should happen in the initializer!

johnpmitsch:
> I think the final thing to do is tie this into the tooling somewhere so this will run automatically. For that I think we can use the `setup-environment` command. It could show the results _and_ write it to a file for later reference (no having to look up the command later, or forgetting to). `setup-environment` is a good place to do this 'cause it's a prerequisite to running the other important commands.
> 
> I'm not sure if there's a non-hacky way to programatically run the Command class. If not, the thing to do would probably be take its behavior (parsing files and messaging) into a service class that will be called by `check-deprecation-warnings` and `setup-environment` command.
> 
> Implementation wise I think there are two options:
> 
> 1. For the given release config (katello, foreman) look at the 'release version' and only check that main repo's deprecations
> 2. For the given release config look at all of the associated repos and check deprecations
> 
> I think the first option is where we should start because we know the version definitively. Sometimes for the other repos we only have branch name and cannot discern the version without more work.

@jturel This seems like a big change from the initial design, I'm not opposed, but I wonder if its better to start as a separate command and entry in the process checklist first? We could vet that it finds what we want, doesn't break every release, and the step makes sense to the release manager. Then could consider making it a more automatic process

lib/tool_belt/commands/check_deprecation_warnings.rb
johnpmitsch: I know its probably invoking some Java flashbacks :smile: but I found "DeprecationWarningFinder" and "DeprecationWarningParser" too similar so tried to distinguish that one finds all deprecation warnings by parsing the code and the other gets you outdated deprecation warnings based on a version. I can brainstorm some other names

jturel:
> @jturel This seems like a big change from the initial design, I'm not opposed, but I wonder if its better to start as a separate command and entry in the process checklist first? We could vet that it finds what we want, doesn't break every release, and the step makes sense to the release manager. Then could consider making it a more automatic process

I thought it was mostly in line with one of your earlier remarks:

> We may be able to parse the repo location and version from the configuration rather than pass it in? Not sure if its possible or one way is preferred over the other, looking for some feedback :)

My thought was 'let's get the foundation completed then ask for the hook into automating' since it's a pretty small tweak code-wise as far as I can tell. Would probably have been better to raise the point earlier, so that's on me.

I think it should definitely be automated soon if not now. I'm not concerned about breaking releases in the sense that I don't think it will break anything. We may actually get more feedback on it because making it more prominent might spark an 'oh, what's this?' reaction.

If we don't automate now then let's make sure it happens by next release cycle IMO!

lib/tool_belt/commands/check_deprecation_warnings.rb
jturel: No big deal if nothing comes to mind. I don't have a better idea either

johnpmitsch:
> > 
> 
> I thought it was mostly in line with one of your earlier remarks:
> 
> > We may be able to parse the repo location and version from the configuration rather than pass it in? Not sure if its possible or one way is preferred over the other, looking for some feedback :)

I can look into using the config with the command in this PR, when I asked this I was still thinking things would be a separate command.

>  I'm not concerned about breaking releases in the sense that I don't think it will break anything.

I mean the deprecation finder tool itself breaking and blocking setup-environment. That is more or less my main concern and reason for keeping things separate at first. I would feel more comfortable seeing it work smoothly in a release or two especially seeing the outcome is the same. I'm open to opinions from other stakeholders though.

jturel:
> I mean the deprecation finder tool itself breaking and blocking setup-environment. That is more or less my main concern and reason for keeping things separate at first. I would feel more comfortable seeing it work smoothly in a release or two especially seeing the outcome is the same. I'm open to opinions from other stakeholders though.

Just clarifying that I wasn't suggesting a blocking approach - just show and write the output.

johnpmitsch:
@jturel updated according to your comments and also updated to use a toolbelt config rather than passing in the repo path and version.

I'm going to keep it as a separate command in this PR as I think that change requires more discussion and buy-in before adding to setup-environment.

Here is example of usage now:
```
[jomitsch@jomitsch tool_belt{dep_warm}]$ ./tools.rb check-deprecation-warnings  configs/katello/3.14.yaml 

Scanning for deprecation warnings in /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/**/*.rb, /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/lib/**/*.rb

Deprecation warnings were found with versions that are outdated, please check below and ensure both the deprecated functionality and the deprecation warning are removed if the deprecation is outdated in this release

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/host_subscriptions_controller.rb line 161 is marked for removal in version 3.5, which is less than or equal to the specified version of 3.14.1.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/repositories_controller.rb line 409 is marked for removal in version 3.3, which is less than or equal to the specified version of 3.14.1.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/content_view_versions_controller.rb line 291 is marked for removal in version 3.4, which is less than or equal to the specified version of 3.14.1.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/v2/subscriptions_controller.rb line 204 is marked for removal in version 2.6, which is less than or equal to the specified version of 3.14.1.

Deprecation warnings were found marked for removal for the *next* release version 3.15. Please check below and file issues aligned to 3.15 to remove the functionality and the deprecation warning. This will ensure the functionality is removed in time for the next release 3.15. The deprecation warning should stay in this release, only file issues for its removal at this time.

Deprecation warning at /home/jomitsch/tool_belt/repos/katello/3.14.1/katello/app/controllers/katello/api/rhsm/candlepin_dynflow_proxy_controller.rb line 70 is marked for removal in version 3.15, which is the *next* release version.

```

johnpmitsch:
@jturel comments have been addressed from what I can tell, so I think its all set to merge :+1:

PR #242, opened 2020-02-19
2.0 client config updates: no fedora28, yes el8

evgeni:


configs/foreman/2.0.yaml
```diff
@@ -408,32 +409,33 @@
...
       build:
         - bash
         - bzip2
+        - centos-release
         - coreutils
         - cpio
         - diffutils
-        - fedora-release
         - findutils
+        - foreman-build
```
ekohl: Do we include this in client repos?

configs/foreman/2.0.yaml
```diff
@@ -408,32 +409,33 @@
...
     external_repos:
-      - fedora-28-updates
-      - fedora-28
+      - centos8-baseos
+      - centos8-appstream
     build_groups:
       build:
         - bash
         - bzip2
+        - centos-release
```
ekohl: @pcreech has suggested in the past that we can use `redhat-release` which both `centos-release` and `fedora-release` provide. Might be good to consider but probably first in nightly.

configs/foreman/2.0.yaml
evgeni: I copied what we have in nightly. My best guess is no, but what do I know ;)

configs/foreman/2.0.yaml
evgeni: Agreed, mind making a card/redmine for that?

configs/foreman/2.0.yaml
evgeni: Would you like to block this until we find out, or postpone that?

ekohl: Not a blocker I think

configs/foreman/2.0.yaml
evgeni: (done)

configs/foreman/2.0.yaml
evgeni: added a card to our backlog

PR #245, opened 2020-02-20
add a note that nightly might got new repos

evgeni:
and these need to be added to the stable config too

### DIFF
--- a/procedures/foreman/branch.md.erb
+++ b/procedures/foreman/branch.md.erb
@@ -52,7 +52,7 @@
 
 # Package build systems
 
-- [ ] Create tool_belt config based on the previous release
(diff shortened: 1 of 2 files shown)


procedures/foreman/branch.md.erb
```diff
@@ -52,7 +52,7 @@
 
 # Package build systems
 
-- [ ] Create tool_belt config based on the previous release
+- [ ] Create tool_belt config based on the previous release and the nightly config which might contain new tags/repos.
```
ekohl:
```suggestion
- [ ] Create tool_belt config based on the previous release and the nightly config which might contain different tags/repos.
```

PR #250, opened 2020-02-25
Add postgersql 12 scldevel to pulp tag

jlsherrill:
for postgresql-debian build

### DIFF
--- a/configs/katello/nightly.yaml
+++ b/configs/katello/nightly.yaml
@@ -134,6 +134,8 @@
       srpm-build:
         - bash
         - python3-rpm-macros
+        - rh-postgresql12-scldevel
+        - scl-utils-build
         - redhat-release
         - redhat-rpm-config
         - rhpkg-simple


ehelms:
That package contains macros needed for the build. As I understand things, that means it has to exist in the buildroot? You can certainly test this with scratch builds @jlsherrill . I really do not see the harm in this though.

ekohl:
I'm worried that we build the ruby postgresql bindings without the SCL but then this part we do build it. I wonder if there will be more macros that are different or could lead to incompatibilities. Mostly afraid of the unknown I guess.

jlsherrill:
@ekohl i think this has to be in the buildroot for the buildrequires to work:

BuildRequires: %{?scl_prefix_postgresql}scldevel
BuildRequires: %{?scl_prefix_postgresql}postgresql-devel


unless you want me to hardcode the scl prefix there?  In which case i might as well hardcode it elsewhere in the spec file.  This seems to be how we do other scl builds.  

Also note that this being added to the pulpcore tag, in isolation from any ruby builds

ekohl:
@jlsherrill can we start with defining that macro in the spec itself to see if that builds? Then we can see about the actual buildroot.

jlsherrill:
@ekohl adding the macro itself doesn't fully do it because i need scl-utils-build in it as well for other macros to be defined properly (added that here).  I have gotten a successful mock build based off that pr and these changes to the mock config:

```
diff --git a/mock/el7-scl.cfg b/mock/el7-scl.cfg
index 5347d0038..6920d66ba 100644
--- a/mock/el7-scl.cfg
+++ b/mock/el7-scl.cfg
@@ -1,7 +1,7 @@
 config_opts['root'] = 'foreman-scl-7-x86_64'
 config_opts['target_arch'] = 'x86_64'
 config_opts['legal_host_arches'] = ('x86_64',)
-config_opts['chroot_setup_cmd'] = 'install @buildsys-build scl-utils-build tfm-build yum vim'
+config_opts['chroot_setup_cmd'] = 'install @buildsys-build scl-utils-build yum vim rh-postgresql12-scldevel'
 config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
 # beware RHEL use 7Server or 7Client
 config_opts['releasever'] = '7'
```

PR #251, opened 2020-02-25
Fix pulpcore inheritance

pcreech:


evgeni:
@pcreech can you also please do the same for 3.15?

PR #254, opened 2020-03-03
check_deprecation_warnings: improve error message

jeremylenz:
Addresses a couple issues I had running `check_deprecation_warnings` locally:

* Improves error messaging if it can't find the repo path.
* Fixes a small bug where it was trying to run a regex `.scan` on something that was expected to be a String, but was not.

### DIFF
--- a/lib/tool_belt/commands/check_deprecation_warnings.rb
+++ b/lib/tool_belt/commands/check_deprecation_warnings.rb
@@ -10,7 +10,9 @@ module ToolBelt
         config = ToolBelt::Config.new(config_file)
         options = config.options
         repo_path = "repos/#{options.namespace}/#{options.project}/"
-        raise "Repo path #{repo_path} not found!" unless File.exist?(repo_path)
+        unless File.exist?(repo_path)
+          raise "Repo path #{repo_path} not found! Did you forget to run './tools.rb setup-environment #{config_file}' ?"
+        end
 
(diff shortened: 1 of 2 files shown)


lib/tool_belt/commands/check_deprecation_warnings.rb
```diff
@@ -10,7 +10,9 @@ def execute
         config = ToolBelt::Config.new(config_file)
         options = config.options
         repo_path = "repos/#{options.namespace}/#{options.project}/"
-        raise "Repo path #{repo_path} not found!" unless File.exist?(repo_path)
+        unless File.exist?(repo_path)
+          raise "Repo path #{repo_path} not found! Did you forget to run './tools.rb setup-environment #{repo_path}' ?"
```
johnpmitsch: `setup-environment` takes a config file, not a repo path. You could specify `config_file` here or keep it generic saying something like "for this release"

johnpmitsch:
Overall looks good and I like the better error message, just a note on the command specified in the message.

lib/tool_belt/commands/check_deprecation_warnings.rb
jeremylenz: @johnpmitsch Thanks, updated!

johnpmitsch:
@jeremylenz I don't have merge permissions, so you will need someone else to do the honors!

PR #256, opened 2020-03-12
Add candlepin for el8 nightlies

zjhuntin:


ehelms:
Is this candlepin tag already in the config for nightly?

zjhuntin:
@ehelms I didn't see anything in tool_belt already

zjhuntin:
Err, does that look better? Or should it be going somewhere else?

configs/katello/nightly.yaml
```diff
@@ -9,6 +9,10 @@
     based_off: null
     arches:
       - x86_64
+  - name: katello-thirdparty-candlepin-el8
+    based_off: null
+    arches:
+      - x86_64
```
ehelms: I think we will need to add `    build_target: katello-nightly-el8-build` as that is how we have EL7 setup. This is not properly captured here in the config. This will mean also adding the katello-nightly-el8 tag setup with this PR.

zjhuntin: ack

ehelms: For EL7, this should actually look like:

```
 :tags:
   - name: katello-thirdparty-candlepin-rhel7
     based_off: null
+    build_target: katello-nightly-rhel7-build
     arches:
       - x86_64
   - name: katello-nightly-rhel7
```

The katello-thirdparty-candlepin-rhel7 tag builds using the katello-nightly-rhel7-build target. The EL8 environment should reflect the same structure.

PR #257, opened 2020-03-17
Add candlepin EL8 mash config and drop thirdparty naming

ehelms:


ehelms:
I was thinking we should drop the `thirdparty` convention and have Candlepin and pulpcore match the same structure as everything else. I can handle the renaming if this is approved.

ekohl:
Mind reflecting this in mashing: https://github.com/theforeman/foreman-infra/blob/b3137be2418bb73244816fd0c989f876a33b5cb9/koji/collection-mash-split.py#L444-L446

PR #258, opened 2020-03-25
document how to update pulp 2to3 migration python package

evgeni:


evgeni:
@jlsherrill @sjha4 :)

sjha4:
Oh wow..Neat! I did something else for this: https://github.com/theforeman/foreman-packaging/pull/4911 . I hope ended up with the same thing. :) Will try this out in a bit..

procedures/pulpcore/update.md.erb
```diff
@@ -12,3 +12,11 @@ podman run --rm --volume $(pwd)/foreman-packaging/:/app/foreman-packaging/:Z  --
 ```
 * [ ] Inspect the generated RPM specs, adjust them if needed
 * [ ] Open a PR with the changes
+
+## Update python-pulp_2to3_migration
+
+The `python-pulp_2to3_migration` package is named differently due to a tito bug (the PyPI name uses dashes, not underscores). Due to this, the regular process of updating it cannot be used.
+
+* [ ] Have a regular `foreman-packaging` clone
+* [ ] Run `obal update python-pulp_2to3_migration -e version=X.Y.Z`
```
ehelms: Let's use them parameters we created `--version`

PR #260, opened 2020-03-26
Updates for Katello 3.15 RC2

jjeffers:


configs/katello/3.15.yaml
```diff
@@ -24,6 +24,7 @@
 :ignores:
   - 28292 # already on the branch
   - 29073 # already on branch
+  - 29360 # rejected
```
jturel: I moved this issue to the recycle bin since it was rejected. No sense in having it triaged to 3.15. So it won't show up! This ignore is unnecessary now

PR #277, opened 2020-05-14
Add creationg of staging repositories to branching

ehelms:


procedures/foreman/branch.md.erb
```diff
@@ -123,6 +123,30 @@ The next step should only be done after all branching, including packaging, has
     - [ ] Remove the oldest version to keep last 3 from the files/directories in the previous steps
 - [ ] Add <%= release %> to [Forklift versions config](https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml)
 
+# Prepare Staging Repositories
```
ekohl, line 126: This now duplicates the Release process. Traditionally this was not needed because we always made sure we could do the release on the same day and you'd move over to the Release procedure. Now we failed on that. I'd actually like to get back into a rhythm where we release RC1 directly after branching.

ehelms: This has some of the same steps as the release process but it does not duplicate the release process as I see it. There can be cases where it as advantageous to prepare the staging repositories while the sources are being prepped. Case in point from 2.1 was that hammer was being finalized and PRs to the 2.1 branch were failing repoclosure. I could have performed this initial signing and prep work to ensure pipelines were working and that repoclosure would pass for updates to the 2.1 branch.

ehelms: I do think branch -> RC1 same day is a great goal. This is just aimed at doing as much preparation work as possible in parallel. To wit, I think helps the stated goal.

ehelms:
@ekohl @tbrisker thoughts?

procedures/foreman/branch.md.erb
```diff
@@ -125,6 +125,30 @@ The next step should only be done after all branching, including packaging, has
...
+  - [ ] [Download](https://github.com/theforeman/theforeman-rel-eng/blob/master/download_rpms)
+  - [ ] [Sign](https://github.com/theforeman/theforeman-rel-eng/blob/master/sign_rpms)
+  - [ ] [Upload RPM signatures](https://github.com/theforeman/theforeman-rel-eng/blob/master/upload_rpm_signatures)
+  - [ ] [Upload RPMs](https://github.com/theforeman/theforeman-rel-eng/blob/master/upload_rpms)
+<% if is_rc %>
+- [ ] Sign "extra" RPMs
+  - [ ] [Download](https://github.com/theforeman/theforeman-rel-eng/blob/master/download_extras)
+  - [ ] [Sign](https://github.com/theforeman/theforeman-rel-eng/blob/master/sign_extras)
+  - [ ] [Upload](https://github.com/theforeman/theforeman-rel-eng/blob/master/upload_extras)
+<% end %>
```
ekohl, line 147: I think we dropped extra by now.

ekohl (commented):
I still think that this is just redundant, especially now that we dropped the extras. Once you get to the point that you branched koji, you might as well move on to the release procedure. Waiting for the actual RC packages is also good because it avoids signing and releasing the nightly builds into a release repo. So IMHO still a :-1: from me.

ehelms:
The original reason for adding this here iis that if you move on to the release procedure and follow it, you will not do the steps added here until at the very end, which means there are a variety of jobs and workflows that can't be tested to ensure they are working ahead of time and has in cases led to delays in the release when there is a break in how the release job runs, or repoclosure, etc.

ekohl:
My primary concern is the nightly builds that would end up in the released repositories. We never purge those on https://yum.theforeman.org/. You can untag them, but until you have the final RCs you would probably not have a passing repoclosure.

PR #279, opened 2020-05-19
fix el8 plugin builds

evgeni:
we need other external repos so that we can build ruby stuff


evgeni:
@pcreech you might hear me screaming over the pond ;-)

PR #282, opened 2020-05-20
adjust katello config to properly fetch build groups

evgeni:
also re-arrange inheritance to match 3.15

### DIFF
--- a/configs/katello/3.16.yaml
+++ b/configs/katello/3.16.yaml
@@ -55,23 +55,22 @@
   - name: katello-3.16-rhel7
     based_off: katello-nightly-rhel7
     helper_tags:
-      katello-3.16-thirdparty-rhel7: null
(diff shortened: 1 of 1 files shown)


evgeni:
(this has been executed to make https://github.com/theforeman/foreman-packaging/pull/5261 work)

configs/katello/3.16.yaml
```diff
@@ -99,45 +98,6 @@
       - test-flat-el8
       - centos8-devel
       - puppetlabs-puppet6-rhel-8
-    build_groups:
```
evgeni, line 102: we set build_package_group_source_tag, no need to add build_groups too

ekohl: Is this something we can do in nightly and foreman as well?

evgeni: foreman yes, nightly no, as that's the source of truth.

PR #285, opened 2020-06-02
document the need to tag foreman-release into el8 client too

evgeni:


evgeni:
this will be needed after https://github.com/theforeman/foreman-packaging/pull/5339

PR #288, opened 2020-06-19
Add pulpcore tag, add support for merge mode

pcreech:


pcreech:
@evgeni @ehelms this could probably use some bikeshedding over the name

configs/pulpcore/nightly.yaml
```diff
@@ -0,0 +1,64 @@
+---
+:project: pulpcore
+:github_org: pulp
```
evgeni, line 3: not sure what implications this has, but pretty sure it won't live in that org ;-)

configs/pulpcore/nightly.yaml
```diff
@@ -0,0 +1,64 @@
...
+    arches:
+      - x86_64
+    inherits:
+      pulpcore-nightly-el8-build:
+        pulpcore-nightly-el8-override: 0
+      pulpcore-nightly-el8-override:
+        pulpcore-nightly-el8: 0
+    external_repos:
+      - name: centos8-baseos
+        mode: bare
```
evgeni, line 23: what is bare, and are we intentionally not using the "flattened" repo?

pcreech: 'bare' is a mode added to koji to support modular repositories.  It tells koji to call mergerepo_c with --pkgorigins, which allows normal mergerepo behaviour, while also providing pkgorigins file for koji.

pcreech: As far as which repo we're using, I structured this around the existing pulpcore tag, which is utilizing the individual repos.  We never converted over to the flattened repo for pulpcore, and we haven't had an issue with it (yet)

configs/pulpcore/nightly.yaml
pcreech: ah, yeah, :D old habits

PR #294, opened 2020-08-05
Add concrete dates to procedures

ehelms:
Adds concrete dates to parts of the tasks, e.g. This mostly works if we always target the same day of the week as the target date, e.g Monday or could adjust for Tuesday. The dates can always be adjusted by hand since this a template.

```
bundle exec ./tools.rb procedure branch foreman 2.2 2020-8-10
```

### DIFF
--- a/lib/tool_belt/commands/procedure.rb
+++ b/lib/tool_belt/commands/procedure.rb
@@ -1,3 +1,5 @@
+require 'date'
+
 module ToolBelt
   module Command
     class ProcedureCommand < Clamp::Command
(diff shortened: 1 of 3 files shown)


lib/tool_belt/commands/procedure.rb
```diff
@@ -24,18 +32,25 @@ def execute
...
           context = {
             is_rc: !extra.nil?,
             extra: extra,
             short_version: "#{major}.#{minor}", # 1.20
             debian_full_version: debian_full_version, # 1.20.0~rc1-1
             full_version: full_version, # 1.20.0-rc1
             version: version, # 1.20.0
+            target_date: parsed_date,
+            two_weeks_before: parsed_date - 14,
+            one_week_before: parsed_date - 7,
```
ekohl, line 53: This appear to be unused

tbrisker (approved):
Thanks @ehelms !

PR #298, opened 2020-08-10
Add Katello 3.17

ehelms:
@chris1984 please to verify

### DIFF
--- a/configs/katello/3.17.yaml
+++ b/configs/katello/3.17.yaml
@@ -1,0 +1,77 @@
+---
+:project: katello
+:github_org: katello
+:releases:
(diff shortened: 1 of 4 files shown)


evgeni:
this is missing EL8 -- intentionally?

ehelms:
We didn't release 3.16 with EL8 support and I don't believe we ewre there yet either with 3.17.

evgeni:
> We didn't release 3.16 with EL8 support and I don't believe we ewre there yet either with 3.17.

Works for me, just was making sure this is not an oversight due to cloning of a wrong template.

PR #302, opened 2020-08-19
move pulpcore to separate, versioned tags

evgeni:
we're not going to have a pulpcore-nightly, as that's just not a thing, so let's shuffle stuff around and have just versioned repos.

I've made 3.4 based off the old katello-pulpcore tag, so that it gets populated properly, and then 3.5 based off 3.4 and 3.6 based off 3.5 which will ensure we have all the deps there to begin with and after tag creation only have to care for pulpcore, it's plugins and any *new* or *updated* dependencies


configs/pulpcore/3.4.yaml
```diff
@@ -0,0 +1,127 @@
...
+:repos: []
+:releases:
+  :3.4:
+    :redmine_version_id: -1
+:mash_scripts:
+  - pulpcore-3.4-el7
+  - pulpcore-3.4-el8
+:tags:
+  - name: pulpcore-3.4-el7
+    based_off: katello-pulpcore-nightly-el7
```
ehelms, line 14: Is that just for this first set of tags to be true?

mash_scripts/pulpcore/3.4/pulpcore-3.4-el7.mash
```diff
@@ -0,0 +1,12 @@
+[pulpcore-3.4-el7]
+rpm_path = %(arch)s/os/Packages
+repodata_path = %(arch)s/os/
+source_path = source/SRPMS
+debuginfo = False
+multilib = False
+multilib_method = devel
+tag = pulpcore-3.4-el7
+inherit = True
+strict_keys = False
```
ehelms, line 10: Implies we are not signing them?

configs/pulpcore/3.4.yaml
evgeni: Yeah, I want to import the current state in nighly as 3.4 and then update from there

mash_scripts/pulpcore/3.4/pulpcore-3.4-el7.mash
evgeni: At the moment correct, I didn't create a key etc yet

PR #305, opened 2020-09-03
Update Transifex in branch docs.

chris1984:
Hit issues with package from yum not working, updated to use pip and added steps on how to install grunt.


jturel (approved):
ACK!

PR #307, opened 2020-09-08
Add ignores for 3.17

chris1984:


chris1984:
@jturel here are the ignores

jturel:
@chris1984 the reason that 3.16.0 issues are showing up is because that release is configured as a prior release to 3.17. The question I have is whether or not 3.16.0 needs to be a prior release to 3.17.0 and the answer to that lies in another question: 'Did 3.17 branch before the 3.16.0 GA gem was created?' You can check, but I'm guessing the answer is probably 'no' which means that all 3.16.0 code is in 3.17.0 already which means it should not be configured as a prior. If this is the case, making that correction will mean a smaller list of ignores, and less confusing cherry-picking overall

chris1984:
@jturel updated

ehelms (approved):

PR #314, opened 2020-11-02
Foreman 2.3 config

pcreech:


tbrisker (approved):
seems reasonable as far as i can tell (hard to properly review a 500+ line yaml)

ekohl (approved):
With CentOS 6 going EOL, I have no issue with dropping that too tbh.

evgeni:
> With CentOS 6 going EOL, I have no issue with dropping that too tbh.

I personally would prefer to keep it around while certain enterprise distros still support it ;)

PR #328, opened 2021-01-19
use correct 8 chars from the 3.9 key

evgeni:
gpg needs the last 8, not the first 8…


ekohl (approved):
Confirmed that this is the right key. In theforeman-rel-eng we have [this](https://github.com/theforeman/theforeman-rel-eng/blob/616ad8891770c3f8f5fd79b736ed52d796dc74f6/settings#L64-L65). Perhaps need a script to print it?

evgeni:
probably, yeah

PR #334, opened 2021-02-01
Add 3.17.3 to prior releases

ianballou:
Eventually we're going to update Katello 4.0, 3.18, and 3.17 to pulp-rpm 3.9.  I'm adding 3.17.3 to 3.18 and 4.0 so that issue propagates. (https://projects.theforeman.org/issues/31785)

### DIFF
--- a/configs/katello/4.0.yaml
+++ b/configs/katello/4.0.yaml
@@ -5,6 +5,8 @@
   :4.0.0:
     :redmine_version_id: 1177
 :prior_releases:
+  :3.17.3:
+    :redmine_version_id: 1365
   :3.18.0:
     :redmine_version_id: 1272
   :3.18.1:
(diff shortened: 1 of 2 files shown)


jlsherrill (approved):

configs/katello/4.0.yaml
```diff
@@ -5,6 +5,14 @@
   :4.0.0:
     :redmine_version_id: 1177
 :prior_releases:
+  :3.17.0:
+    :redmine_version_id: 1282
+  :3.17.1:
+    :redmine_version_id: 1337
+  :3.17.2:
+    :redmine_version_id: 1342
+  :3.17.3:
```
jturel, line 14: Everything before 3.17.3 has gone GA by the time 4.0 branches so I don't think those are needed here? +1 to 3.17.3 though

ianballou: oops, you're right

ianballou: All fixed now.

jturel (approved):

PR #336, opened 2021-02-17
Updates config for Katello 4.0.0.rc2

jjeffers:


configs/katello/4.0.yaml
```diff
@@ -36,7 +36,8 @@
   :foreman_virt_who_configure:
     :branch: 0.5.5
     :repo: https://github.com/theforeman/foreman_virt_who_configure.git
-:ignores: []
+:ignores:
+  - 31308 # already in
```
jturel, line 40: It looks like the related PR has a mismatch with the issue in the PR vs what was in the commit. I wonder if correcting the PR name will fix it ?

jjeffers: It's this commit https://github.com/Katello/katello/pull/9035/commits/93b9558453b99abea6322a9a2741a1c50634128c

jturel: Ah, the commit itself is mismatched with the issue name of the PR -> https://github.com/Katello/katello/pull/9035/commits

Nothing we can do at this point - so I agree w/ the ignore!

PR #337, opened 2021-02-24
Update the tagging procedure

ekohl:
This uses more scripts in theforeman-rel-eng. Perhaps a new script should be introduced to push the tags too.

I also noticed I needed to symlink foreman-proxy to smart-proxy. It's very much a worksforme PR, but I think it can further unify our process by keeping all tooling in a single repository.


ekohl:
This was done in 6296caf02239561ed6f30451de68bd1ff0edd2bb

PR #340, opened 2021-03-15
Katello 4.0.0-rc3 updates

jjeffers:


configs/katello/4.0.yaml
```diff
@@ -38,6 +38,9 @@
     :repo: https://github.com/theforeman/foreman_virt_who_configure.git
 :ignores:
   - 31308 # already in
+  - 31967 # for katello 3.18
+  - 31636 # migration fixes for katello 3.18
+  - 31712 # not yet included in Foreman 2.4-stable
```
jturel, line 43: Maybe this should be triaged to 4.1? The foreman issue isn't triaged for 2.4 -> https://projects.theforeman.org/issues/31629

Doing this implies you can remove the ignore here

jjeffers: @jturel I somehow missed this. I can update the issue, if need be.

jturel:
3.18.3 has been added in Redmine since this was opened - that should be among the priors too!

configs/katello/4.0.yaml
jturel: Yeah I think it should be 4.1

jturel (approved):

PR #346, opened 2021-04-08
Create an artifacts job

jjeffers:
This adds a workflow where the last 3 Katello versions changelogs and cherry-picks are generated and uploaded as an artifact.


jjeffers:
The actions did not complete successfully here, but they did on my fork... I'll try and see why it is different.

See: https://github.com/jjeffers/tool_belt/actions/runs/730725390

jjeffers:
I think the failures are due to not having the repository secret for GH_ACCESS_TOKEN, which is exported in the workflows as ENV variable GITHUB_ACCESS_TOKEN.

ehelms:
What is the intended workflow, as a maintainer, to trigger the run of this and creation of the artifacts?

jjeffers:
> What is the intended workflow, as a maintainer, to trigger the run of this and creation of the artifacts?

Ideally, every configuration change would generate an updated set of CHANGELOG.md and cherry-pick summaries.

jturel:
There is not always a config update that coincides with the need to generate a new changelog or set of cherry-picks. And not every config update means that a new changelog or cherry-pick report should be generated.

.github/workflows/secondary.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+    - uses: actions/checkout@v2
+    - id: set-matrix
+      run: echo "::set-output name=matrix::$(ls configs/katello/ | sed '/-/!{s/$/_/}' | sort -rV | sed 's/_$//' | grep yaml | grep -v nightly | head -n 3 | jq -cRs 'split("\n")[:-1]')"
```
ekohl, line 13: I'm not sure what you're exactly after, but how about using this as a base?
```console
$ ls configs/katello | xargs -I '{}' basename '{}' .yaml | sort -n | tail -n 3
3.8
3.9
4.0
```

jjeffers:
> There is not always a config update that coincides with the need to generate a new changelog or set of cherry-picks. And not every config update means that a new changelog or cherry-pick report should be generated.

I agree that most changes don't need a newly generated set of reports, etc. I am more concerned with eliminating as many manual steps as possible in the release process. So for the relatively few instances where we are concerned or interested in these artifacts, we have a consistent, automatic generation. Hopefully, this helps to eliminate errors when folks have to do these steps by hand.

.github/workflows/secondary.yml
ekohl: Or even better: use `basename --suffix=.yaml configs/katello/*.yaml | sort -n | tail -n 3`

jturel:
> > There is not always a config update that coincides with the need to generate a new changelog or set of cherry-picks. And not every config update means that a new changelog or cherry-pick report should be generated.
> 
> I agree that most changes don't need a newly generated set of reports, etc. I am more concerned with eliminating as many manual steps as possible in the release process. So for the relatively few instances where we are concerned or interested in these artifacts, we have a consistent, automatic generation. Hopefully, this helps to eliminate errors when folks have to do these steps by hand.

Definitely useful to automate this. I think it represents a starting point for something that we'll have to see how it plays out. Then see if there's anything we can do to get more mileage from it. For example it'd be nice to have the changelog regenerated when merging cherry picks in to one of the release branches or something like that (separate thing from this PR obviously)

.github/workflows/secondary.yml
jjeffers: Ah, ok yes I prefer a more compact expression - but in this case I just wanted the 3 most recent versions, "4.0, 3.18, 3.17", for example. The default sorting doesn't get me there.

ekohl: I thought that's what `sort -n` does. From `man sort`:

```
-n, --numeric-sort
              compare according to string numerical value
```

Looks like we should use `-V`:
```
-V, --version-sort
              natural sort of (version) numbers within text
```

So then it becomes:
```console
$ basename --suffix=.yaml configs/katello/*.yaml | grep -v nightly | sort -V | tail -n 3
3.17
3.18
4.0
```

ekohl: And I now see you already found that.

.github/workflows/secondary.yml
```diff
@@ -0,0 +1,40 @@
...
+          ruby-version: '2.5'
+          bundler-cache: true
+      - name: setup environment
+        run: bundle exec ./tools.rb setup-environment configs/katello/${{ matrix.version }}
+      - name: generate CHANGELOG
+        run: bundle exec ./tools.rb changelog configs/katello/${{ matrix.version }}
+      - name: generate cherry-picks
+        env:
+          GITHUB_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+        run: bundle exec ./tools.rb cherry-picks configs/katello/${{ matrix.version }}
```
ekohl, line 34:
```suggestion
        run: bundle exec ./tools.rb cherry-picks configs/katello/${{ matrix.version }}.yaml
```

ekohl:
Closing for inactivity. This does have potential, but I'd suggest to take it a step further and do it in the repository itself and create the PR(s). In https://github.com/theforeman/foreman-packaging/blob/master/.github/workflows/bump_packages.yml we do that.

PR #347, opened 2021-04-09
Move foreman installer steps up 2 weeks

jjeffers:


procedures/katello/branch.md.erb
```diff
@@ -29,6 +29,12 @@
 ## Release Packager
 
 - [ ] Ensure tool_belt config is merged and output from `./tools.rb koji configs/katello/<%= release %>.yaml` matches expectations
+- [ ] Ensure releases of installer modules, this is typically done as part of Foreman module release workflow
+    - [ ] [candlepin](https://github.com/theforeman/puppet-candlepin)
+    - [ ] [pulp](https://github.com/theforeman/puppet-pulp)
```
ekohl, line 34: This has been dropped since 4.x

procedures/katello/branch.md.erb
```diff
@@ -29,6 +29,12 @@
...
+    - [ ] [pulpcore](https://github.com/theforeman/puppet-pulpcore)
+    - [ ] [foreman_proxy_content](https://github.com/theforeman/puppet-foreman_proxy_content)
+    - [ ] [katello](https://github.com/theforeman/puppet-katello)
```
ekohl, line 37: Do we need to name every one? We do it as part of the Foreman Installer release so really the message should be that all (major) changes should be in 2 weeks before the branching happens.

jjeffers: AFAIK there is no reason to list each one. Thanks!

procedures/katello/branch.md.erb
```diff
@@ -29,12 +29,7 @@
 ## Release Packager
 
 - [ ] Ensure tool_belt config is merged and output from `./tools.rb koji configs/katello/<%= release %>.yaml` matches expectations
-- [ ] Ensure releases of installer modules, this is typically done as part of Foreman module release workflow
-    - [ ] [candlepin](https://github.com/theforeman/puppet-candlepin)
-    - [ ] [pulp](https://github.com/theforeman/puppet-pulp)
-    - [ ] [pulpcore](https://github.com/theforeman/puppet-pulpcore)
-    - [ ] [foreman_proxy_content](https://github.com/theforeman/puppet-foreman_proxy_content)
-    - [ ] [katello](https://github.com/theforeman/puppet-katello)
+- [ ] Ensure major changes to Foreman installer modules are completed
```
ekohl, line 32: Is it the task of the release packager? I think the release owner is in charge of making sure features are completed. IMHO the goal should be to make sure during stabilization we are really working on making it stable, not finishing features which can destabilize it.

jjeffers: @jturel @jlsherrill thought's about this?

jturel: I think the bullet for the packager would be something like "confirm with release owner that all expected changes are present before performing an installer modules release" - something to ensure that the two are in sync. However it seems rare to me that Katello devs are even slightly aware of the installer changes going in to a release that they didn't ask for themselves so 'expected' is a vague term in my example.

ekohl: That's sort of my point: often a change is made in Katello somewhere and then it turns out to need an installer change. Often that happens during or even after stabilization. My goal is to prevent that. IMHO it's the release manager's job to ensure those changes are in before Foreman stabilizes. The Foreman packager should take care of foreman-installer which means the Katello packager doesn't really have to do anything.

jturel: I remember one such case not too long ago in fact. 

@jjeffers it sounds like this change becomes:

 - leave the block which you deleted
 - add a new point for the release owner 2 weeks prior to branching to work with the team to ensure all installer changes are done

procedures/katello/branch.md.erb
```diff
@@ -25,10 +25,16 @@
 - [ ] Ensure stable Pulp release
   - [ ] Update the [katello-repos](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello-repos/katello-repos.spec#L1-L2)
   - [ ] Update [forklift](https://github.com/theforeman/forklift/blob/master/roles/katello_repositories/defaults/main.yml#L4-L5)
+- [ ] Coordinate with Release Packager that expected changes to associated installer modules are completed.
 
 ## Release Packager
 
 - [ ] Ensure tool_belt config is merged and output from `./tools.rb koji configs/katello/<%= release %>.yaml` matches expectations
+- [ ] Confirm with release owner that all expected changes are present before performing an installer modules release.
```
ekohl, line 33: This gives the impression that the Katello Release Packager does these releases but in practice they don't. Until now that has **always** been done by the installer maintainers. I doubt release packagers even have the permissions to do so.

ekohl (approved):
I'm merging this now since it's an improvement that looks correct to me and we can iterate later.

PR #351, opened 2021-04-22
Update puppet modules in foreman branching procedure

wbclark:
Pulp 2 has been dropped and Foreman's Puppetserver has been
split into a separate module


ekohl (approved):
Thanks!

PR #353, opened 2021-05-03
Add foreman 2.5 config and mash scripts

zjhuntin:


configs/foreman/2.5.yaml
```diff
@@ -0,0 +1,541 @@
+---
+:project: foreman
+:github_org: theforeman
+:releases:
+  :2.5.0:
+    :redmine_version_id: 1319
```
evgeni, line 6:
```suggestion
    :redmine_version_id: 1367
```

(not sure it's used anywhere, but yeah)

evgeni (changes requested):
one nit

configs/foreman/2.5.yaml
zjhuntin: Thanks!

PR #357, opened 2021-05-05
switch 2.5 and nightly el7 config to rh-ruby27

evgeni:


configs/foreman/nightly.yaml
```diff
@@ -105,7 +105,7 @@
         - make
         - patch
         - redhat-rpm-config
-        - rh-ruby25-scldevel
+        - rh-ruby27-scldevel
```
evgeni, line 108: I have no idea why this was still on 25..

pcreech (approved):
ACK

PR #358, opened 2021-05-05
Sync up Foreman nightly tags build groups

ehelms:


configs/foreman/nightly.yaml
```diff
@@ -570,6 +567,7 @@
       srpm-build:
         - bash
         - centos-release
+        - redhat-release-server
```
evgeni, line 570: while I find seeing rh-rel-server weird in this list, here centos is not removed?

ehelms: Yea, its a good callout. I'm just the messenger, don't throw :egg: :egg: :egg:

evgeni: Easter is over anyways :rabbit:

evgeni (approved):
given that reflects the status quo: ack

but we shall clean this up.

PR #359, opened 2021-05-05
Add katello 4.1 mash scripts

zjhuntin:


evgeni (approved):

mash_scripts/katello/4.1.0/katello-4.1-el8.mash
```diff
@@ -0,0 +1,12 @@
+[katello-4.1-el8]
```
ekohl, line 1: Why 4.1.0 in the directory path? All others are x.y without .z.

ehelms: Look again, they are all x.y.z which matches what the code auto-generates.

ekohl: Hmm, right. Except 3.3. I do feel like we should move to x.y for simplicity and consistency with https://github.com/theforeman/tool_belt/tree/master/configs/katello though.

PR #375, opened 2021-06-03
Switch from centos7-os to centos-7-server external repositories

ehelms:
Koji has duplicate external repositories for CentOS 7 Server and
Updates and this consolidates the use.

### DIFF
--- a/configs/katello/3.15.yaml
+++ b/configs/katello/3.15.yaml
@@ -6,7 +6,7 @@
   :3.15.3:
     :redmine_version_id: 1257
   :3.15.2:
-    :redmine_version_id: 1238 
+    :redmine_version_id: 1238
   :3.15.1:
     :redmine_version_id: 1211
   :3.15.0:
@@ -80,5 +80,5 @@
         katello-pulpcore-3.15-el7: 0
     build_package_group_source_tag: katello-pulpcore-nightly-el7-build
     external_repos:
-      - centos7-os
-      - centos7-updates
+      - centos-7-server
+      - centos-7-server-updates
--- a/configs/katello/3.16.yaml
+++ b/configs/katello/3.16.yaml
@@ -2,7 +2,7 @@
 :project: katello
 :github_org: katello
 :releases:
-  :3.16.2: 
+  :3.16.2:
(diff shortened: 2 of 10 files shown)


ehelms:
Needs this change https://github.com/theforeman/foreman-infra/pull/1614/files#diff-5785de757e436cf61a963a490cdad9051c83b296873434569fc8a2733406042fR16 to be able to merge so that the tags when updated will behave the same.

configs/katello/3.15.yaml
```diff
@@ -80,5 +80,5 @@
         katello-pulpcore-3.15-el7: 0
     build_package_group_source_tag: katello-pulpcore-nightly-el7-build
     external_repos:
-      - centos7-os
-      - centos7-updates
+      - centos-7-server
```
ekohl, line 83: Can we standardize on `centos-7-os` instead? That's the repo name that CentOS actually uses (as well as updates). There is no such thing as CentOS Server.

ehelms: I can do that as a follow up step, once everything is using a single repository, renames become easier.

ekohl: If anything, I think we should remove `centos-7-server` then. Note you are renaming updates as well here by adding a dash.

ehelms: Right now, I want to aim for alignment first (and removing duplication). Then do mass updates. The change to remove `centos-7-server` would at present touch 187 tags:

```
$ grep -r centos-7-server configs/ | wc -l
187
```

Now we wont actually need to update all of them but I want to give an indicator of my preferred strategy here.

configs/katello/3.15.yaml
```diff
@@ -80,5 +80,5 @@
...
+      - centos-7-updates
```
ekohl, line 84: Today this doesn't exist. Do you intend to rename it?

ehelms: Vim mass update fail, this is actually named `centos-7-server-updates` and should now be reflected.

ekohl:
@ehelms given we're moving to Copr I'm going to close this.

PR #377, opened 2021-06-11
pulpcore 3.13

evgeni:


pcreech:
Commit message for the conifg says "add pulpcore 3.11 config", when the file it's adding is 3.13.

But content looks correct.

evgeni:
fixed :)

PR #378, opened 2021-06-15
Declare explicit roles in procedures

ekohl:
In Foreman Release there were no explicit roles so those were added.

It also aligns Foreman and Katello naming for the Release Engineer role.

This implements point 1 and 3 as laid out in the [communication RFC](https://community.theforeman.org/t/branching-release-process-communication/23965). The second point is something humans should adopt.

Also includes an unrelated fix to use the tarballs_release script.

### DIFF
--- a/procedures/foreman/branch.md.erb
+++ b/procedures/foreman/branch.md.erb
@@ -1,5 +1,11 @@
 [ ] Make this post a wiki
 
+## Roles
+
+* Release Owner: <%= owner %>
+* Release Engineer: <%= engineer %>
+* Installer Maintainer: @
+
 # Prep Week: <%= two_weeks_before %> to <%= two_weeks_before + 4 %>
 
 ## Installer Maintainer
--- a/procedures/foreman/release.md.erb
+++ b/procedures/foreman/release.md.erb
@@ -1,6 +1,14 @@
 [ ] Make this post a wiki
 
+## Roles
+
+* Release Owner: <%= owner %>
+* Release Engineer: <%= engineer %>
+* Installer Maintainer: @
+
 # Manual updates: <%= target_date %>
+
+## Release Engineer
 
 - [ ] Update manual if applicable for any additional installation steps
 - Update [release notes section](gh-pages/_includes/manuals/<%= short_version %>/1.2_release_notes.md) in the manual:
@@ -17,8 +25,13 @@
 
 # Preparing code: <%= target_date %>
 
+## Installer Maintainer
+
 - [ ] Make patch releases of installer modules that have important changes
   - [ ] Branch to MAJ.MIN-stable if recent changes to the module aren't suitable for patch (x.y.z) release
+
+## Release Owner
+
 <% unless (is_rc || full_version.end_with?('0')) %>- [ ] Add a new [Redmine version](https://projects.theforeman.org/projects/foreman/settings/versions) for the next minor, unless the series is EOL. Be sure the version is set to sharing with subprojects.
 <% end %>- [ ] Remove/change target version field for any open Redmine tickets assigned to the release still (next minor, unset it or reject)
 - [ ] Ensure that code in git matches issues fixed in <%= full_version %> in redmine. [issues.rb](https://gist.github.com/tbrisker/3320097f1d8c9abfb2dbcf8d8e216bba) can be used to generate a comparison between the two.
@@ -26,6 +39,8 @@
 <% end %>
 
 # Tagging a release: <%= target_date %>
+
+## Release Owner
 
 - In foreman <%= short_version %>-stable:
   - [ ] Make sure [test_<%= short_version.tr('.', '_') %>_stable](https://ci.theforeman.org/job/test_<%= short_version.tr('.', '_') %>_stable/) is green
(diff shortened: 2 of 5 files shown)


procedures/foreman/branch.md.erb
```diff
@@ -1,5 +1,11 @@
 [ ] Make this post a wiki
 
+## Roles
+
+* Release Owner: <%= owner %>
+* Release Engineer: <%= engineer %>
+* Installer Maintainer: @
```
evgeni, line 7: Is this intentionally left empty?

procedures/foreman/release.md.erb
```diff
@@ -1,7 +1,15 @@
...
 
+## Roles
+
+* Release Owner: <%= owner %>
+* Release Engineer: <%= engineer %>
+* Installer Maintainer: @
+
 # Manual updates: <%= target_date %>
 
+## Release Engineer
```
evgeni, line 11: Should these headers repeat the username?

procedures/foreman/release.md.erb
```diff
@@ -38,8 +53,11 @@
   - [ ] Tag the release using [tag.sh](https://gist.github.com/tbrisker/d51cedea51ccbcca00e72c1fc61550fe) `tag.sh <%= full_version %> && git push upstream <%= short_version %>-stable --follow-tags`
 - In foreman-installer <%= short_version %>-stable:
   - [ ] Tag the release using [tag.sh](https://gist.github.com/tbrisker/d51cedea51ccbcca00e72c1fc61550fe) `tag.sh <%= full_version %> && git push upstream <%= short_version %>-stable --follow-tags`
-- [ ] Run the Jenkins [Tarballs Release](https://ci.theforeman.org/job/tarballs-release/) to create tarballs
+
+## Release Engineer
+
 - [ ] Update release version similar to [here](https://github.com/theforeman/theforeman-rel-eng/commit/2029a9688da00d9c385c3438dd71b594ba5f728e)
+- [ ] Run the Jenkins [Tarballs Release](https://ci.theforeman.org/job/tarballs-release/) to create tarballs using [release_tarballs](https://github.com/theforeman/theforeman-rel-eng/blob/master/release_tarballs)
```
evgeni, line 60: Why moving this off to rel-eng? This step happens directly after tagging, and could be already finished the moment rel-eng sees the ping if started by rel-owner (as it is right now).

No hard blocker, more curious why :)

evgeni (commented):
Comments/questions, but full ack on the explicit documentation of the ownership.

Aligns well with the changes @pcreech @zjhuntin and @odilhao been planning, where we assign an rel-eng person to the release, instead of switching if the release happens to span multiple sprints.

procedures/foreman/branch.md.erb
ekohl: Yes, it was sort of a one-off and I wasn't entirely sure if we should add it to the branch command.

procedures/foreman/release.md.erb
ekohl: Interesting idea.

procedures/foreman/release.md.erb
ekohl: I sort of assumed it was done by rel-eng. Also there's a script in theforeman-rel-eng for it.

evgeni: So far it was always @tbrisker who kicked that off. No idea if by script or not.

tbrisker: I manually trigger it when I'm done tagging. I guess I see it as the last step in getting to "Here are the artifacts you need, now build them into packages please", but it could also be done by rel-eng just as well when pinged.

ekohl: Perhaps we should also make sure the release owner can use the rel-eng scripts to make this easier? There is also [tag_project](https://github.com/theforeman/theforeman-rel-eng/blob/master/tag_project) which we today document as separate steps for example.

My idea is that we can further smooth the process: rather than long instructions just point to scripts. At that point, it matters less who does it.

Another idea: verifying if Jenkins jobs are green can also be done in a script and the exit code can indicate if they are.

In practice what you can already do is:
```
./release_tarballs && ./download_tarballs && ./inspect_tarballs
```

I also think we should also qualify the inspect step. What I always manually did was verify the version is correct in VERSION. Perhaps we can automate that too using tar and grep. It could also be a place to prevent supply chain attacks (if Jenkins would be compromised) but I'm not sure how we would achieve that.

However, maybe we're getting a bit off-topic here.

ekohl: Oh and to be clear, the real reason I moved it down was to put it below the updating of theforeman-rel-eng to be able to use the script and I considered updating that part of the release engineer's role.

tbrisker: In general I'm all for automating as much as possible. I think the key points that need a human to do something (be it run a script or trigger a job) are:
- The release is tagged and ready for building, start the train
- Verify that this is the real and correct files/code/package/whatever and sign everything. Ideally this could also be automated but if we're thinking of supply chain security we probably can't trust jenkins to verify the code is correct or hold the keys to signing, that would have to be done on some separate, hopefully secure system.

Anything else could potentially be done by a machine, especially if all the human right now does is run a script. Let me know if i'm missing some step that is crucial to have human involvement in.

ekohl: I've removed this change and limited it to just adding roles. I can open a separate PR for more scripting.

tbrisker (approved):
Thanks @ekohl !

PR #382, opened 2021-07-19
Only release on days with a working day after

ekohl:
It is considered good practice to release on a day when there is a working day following it. The risk is that something does break and if you have to wait a whole weekend to fix it, it's bad. Note that goes not only for users but also developers.

This is all not relevant for release candidates. Everyone should be aware that release candidates are not meant for production and the risks associated with them.


upadhyeammit (approved):
LGTM!

ehelms (approved):

PR #389, opened 2021-08-06
Switch nightly to use Puppet 7 external repository

ehelms:


configs/katello/nightly.yaml
```diff
@@ -65,7 +65,6 @@
       - flat-ruby27-el8
       - test-flat-el8
       - centos8-devel
-      - puppetlabs-puppet6-rhel-8
```
ekohl, line 68: I assume this is removed since it doesn't actually need Puppet, right?

ehelms: Yea, I cannot see a reason it would need it.

pcreech (approved):
ACK

PR #391, opened 2021-08-11
Pulpcore 3.15

pcreech:


pcreech:
Depends on https://github.com/theforeman/foreman-infra/pull/1635/files

Odilhao (approved):

configs/pulpcore/3.15.yaml
```diff
@@ -0,0 +1,127 @@
...
+        - python3-rpm-macros
+        - rh-postgresql12-scldevel
+        - scl-utils-build
+        - redhat-release
+        - redhat-rpm-config
+        - rhpkg-simple
+        - rpm-build
+        - shadow-utils
+  - name: pulpcore-3.15-el8
+    based_off: pulpcore-3.13-el8
```
evgeni, line 70: Nope

configs/pulpcore/3.15.yaml
```diff
@@ -0,0 +1,127 @@
...
+  :3.15:
+    :redmine_version_id: -1
+:mash_scripts:
+  - pulpcore-3.15-el7
+  - pulpcore-3.15-el8
+:strict_keys: true
+:gpg_key: 6D05E64C
+:tags:
+  - name: pulpcore-3.15-el7
+    based_off: pulpcore-3.13-el7
```
evgeni, line 16: Nope

evgeni (changes requested):

configs/pulpcore/3.15.yaml
pcreech: indeed.... 
fixed

configs/pulpcore/3.15.yaml
```diff
@@ -0,0 +1,127 @@
...
+        pulpcore-3.15-el7-override: 0
+      pulpcore-3.15-el7-override:
+        pulpcore-3.15-el7: 0
+    external_repos:
+      - centos7-os
+      - centos7-updates
+      - centos-sclo-rh-rhel-7
+      - epel-7
+    build_groups:
+      build:
```
pcreech, line 33: And, something I missed, need to add rh-python38 into the el7 groups

configs/pulpcore/3.15.yaml
```diff
@@ -0,0 +1,130 @@
...
+        - grep
+        - gzip
+        - info
+        - make
+        - patch
+        - python3-rpm-macros
+        - redhat-release
+        - redhat-rpm-config
+        - rpm-build
+        - rh-python38-scldevel
```
ehelms, line 52: I think this should be `rh-python38-build`

configs/pulpcore/3.15.yaml
```diff
@@ -0,0 +1,130 @@
...
+  :3.15:
+    :redmine_version_id: -1
+:mash_scripts:
+  - pulpcore-3.15-el7
+  - pulpcore-3.15-el8
+:strict_keys: true
+:gpg_key: 6D05E64C
+:tags:
+  - name: pulpcore-3.15-el7
+    based_off: pulpcore-3.14-el7
```
ehelms, line 16: Given these configs define the build groups, and external repositories, is the based off even needed?

evgeni: yes, otherwise the deps won't get tagged.

granted, in the case of 3.15, it's not that important, as we need to rebuild everything…

configs/pulpcore/3.15.yaml
```diff
@@ -0,0 +1,130 @@
...
+:nightly: false
+:repos: []
+:releases:
+  :3.15:
+    :redmine_version_id: -1
+:mash_scripts:
+  - pulpcore-3.15-el7
+  - pulpcore-3.15-el8
+:strict_keys: true
+:gpg_key: 6D05E64C
```
evgeni, line 13:
```suggestion
:gpg_key: F7D3B452
```

pcreech: Done

configs/pulpcore/3.15.yaml
evgeni: We're going to create a `tfm-pulpcore` SCL, and for Foreman we include `rh-ruby27-scldevel` and `tfm-build`, so I am going to assume we would want `rh-python38-scldevel` and `tfm-pulpcore-build`.

Given the later doesn't exist (yet), I am going to merge and run this as-is, then build the SCL package, and then add it here.

evgeni (approved):

PR #396, opened 2021-09-01
use python38-rpm-macros to build against python38

evgeni:


Odilhao (approved):

evgeni:
> We have @theforeman/packaging. Should add this repo to that group?

Either works for me, yeah!

ehelms:
> > We have @theforeman/packaging. Should add this repo to that group?
> 
> Either works for me, yeah!

Done.

PR #402, opened 2021-09-09
Katello 3.18.5 config update

jturel:


ekohl (approved):
We should really update tool_belt to query https://projects.theforeman.org/projects/katello/versions.json for this. Manually mapping IDs feels like a bad practice.

PR #403, opened 2021-09-21
Katello 4.0.3 udpates

jjeffers:


configs/katello/4.0.yaml
```diff
@@ -19,6 +21,8 @@
     :redmine_version_id: 1370
   :3.18.4:
     :redmine_version_id: 1396
+  :3.18.5:
```
jturel, line 24: I think everything on 3.18.5 was specific to the KATELLO-3.18 branch ie it's not an actual prior - unless your cherry-picks actually contained something from 3.18.5?

jjeffers: I think that everything indicated for picks were ignored because they were migration related.

jturel: Ack. Can you remove 3.18.5 then?

configs/katello/4.0.yaml
```diff
@@ -72,6 +75,17 @@
   - 32786 # for migration/3.18
   - 32887 # for migration/3.18
   - 32888 # for migration/3.18
+  - 33131 # for migration/3.18
```
jturel, line 78: Since 3.18.5 any new ignores from that release can be removed here

jturel (approved):
ty james!

PR #404, opened 2021-09-22
Add a testweek RC1 note

lzap:
SSIA


ekohl (commented):
Thinking out loud, should we make a `testweek.md.erb` in https://github.com/theforeman/tool_belt/tree/master/procedures/foreman with supporting code in https://github.com/theforeman/tool_belt/blob/master/lib/tool_belt/commands/procedure.rb? Then the release owner can easily post this as part of the procedure.

cc @upadhyeammit

upadhyeammit:
> Thinking out loud, should we make a `testweek.md.erb` in https://github.com/theforeman/tool_belt/tree/master/procedures/foreman with supporting code in https://github.com/theforeman/tool_belt/blob/master/lib/tool_belt/commands/procedure.rb? Then the release owner can easily post this as part of the procedure.
> 
> cc @upadhyeammit 

Agree! Release process has more eyes and even if release owner is not available due to any reasons there is always some backup who completes the checklist.

lzap:
I spent some time trying to understand what this does, if you don't mind I would rather be just poked and continue organizing the event.

ekohl:
I've opened https://github.com/theforeman/tool_belt/pull/418 which implements it as a procedure.

PR #405, opened 2021-09-28
add ruby and swig modules to pulpcore buildroot

evgeni:
libsolv needs these to build


evgeni:
swig repo creation is in https://github.com/theforeman/foreman-infra/pull/1655

pcreech (approved):

PR #411, opened 2021-11-09
Configs and Mash for Foreman 3.1

Odilhao:


configs/foreman/3.1.yaml
```diff
@@ -0,0 +1,492 @@
+---
+:project: foreman
+:github_org: theforeman
+:releases:
+  :3.1.0:
+    :redmine_version_id: 1408
```
ekohl, line 6:
```suggestion
    :redmine_version_id: 1488
```

ekohl (changes requested):

configs/foreman/3.1.yaml
Odilhao: Thanks, where I can find the ID for Redmine? I tried to find it yesterday.

Odilhao: Done, it's now
>  :redmine_version_id: 1488

ekohl: I took it from the URL on https://projects.theforeman.org/projects/foreman/roadmap. Really it should use the API which can tell you the ID but that's some coding that needs to be done in tool belt itself.

upadhyeammit:
@ekohl if there isn't any more changes can we merge this?

ekohl (commented):
The only thing I wonder about is EL6 client support. We dropped it in 3.0 so it's surprising to introduce it again.

Odilhao:
> The only thing I wonder about is EL6 client support. We dropped it in 3.0 so it's surprising to introduce it again.

We still have EL6 at our nightly definitions, we probably want to remove from there as well

upadhyeammit:
> > The only thing I wonder about is EL6 client support. We dropped it in 3.0 so it's surprising to introduce it again.
> 
> We still have EL6 at our nightly definitions, we probably want to remove from there as well

For nightly client I can see el5, el6 and FC27, FC28, FC29 . If we dont have any specific purpose to keep it with nightly only then all of these should be removed as well?

Odilhao:
Dropped both F29 and EL6 from nightly and also removed from the EL6 mash from 3.1

configs/foreman/nightly.yaml
```diff
@@ -361,56 +359,6 @@
       foreman-plugins-nightly-rhel7-dist:
         foreman-plugins-nightly-rhel7: 2
         foreman-plugins-nightly-nonscl-rhel7: 0
-  - name: foreman-client-nightly-rhel6
```
ekohl, line 364: This would probably also need to be reflected in foreman-packaging so it won't attempt to build anymore.

configs/foreman/nightly.yaml
```diff
@@ -518,56 +466,6 @@
         - rhpkg-simple
         - rpm-build
         - shadow-utils
-  - name: foreman-client-nightly-fedora29
```
ekohl, line 521: We kept F29 in to make it easy to update it to current Fedora. Are we giving up on Fedora now? If so, that also needs to be reflected in foreman-packaging.

Odilhao: I removed the changes on nightly, we can discuss this in another PR I think. Right now I'm only sending changes for 3.1

upadhyeammit:
@ekohl as discussed not dropping anything from nightly in this pr, can you check once if we are good to merge this?

ekohl (approved):

ekohl (approved):
Also deployed now.

PR #427, opened 2022-02-09
Add Katello branching procedure steps for Katello Next

ianballou:
In order to better sync-up with Foreman release practices, Katello will wait until branching to create the latest release in redmine.  If, say, Katello 4.5.0 is created in redmine before Katello 4.4.0 has branched, the "fixed in version" field will incorrectly be set by prprocessor to 4.5.0.  To work around this, all future bugs that should be on a version after the latest pre-branched version will be triaged to "Katello Next".

For now, it will be up to the release owner to update Katello Next and related redmine releases at branching time.  The new instructions below illustrate this.

@ekohl wrote up some more detailed notes here: https://community.theforeman.org/t/release-team-meeting-agenda-2022-02-09/27243/3

### DIFF
--- a/procedures/katello/branch.md.erb
+++ b/procedures/katello/branch.md.erb
@@ -90,6 +90,12 @@
       cp -rf public/apipie-cache/* $GITDIR/theforeman.org/plugins/katello/$VERSION/api
       find $GITDIR/theforeman.org/plugins/katello/$VERSION/api -name "*.json" -type f -delete
       sed -i "/layout: /d" $GITDIR/theforeman.org/plugins/katello/$VERSION/api/index.md
+- [ ] Prepare "Katello Next" and future redmine versions
+  - [ ] Rename the "Katello Next" release to Katello <%= develop %>.0
+  - [ ] Recreate the "Katello Next" release and indicate that it is a placeholder for issues belonging to the next version of Katello
+  - [ ] Create the next 'z' release: Katello <%= release %>.1
+  - [ ] Create the "Katello <%= release %>.1 TODO" custom query using the last one as a template
+
 
 ## Release Engineer
 


procedures/katello/branch.md.erb
```diff
@@ -90,6 +90,12 @@
       cp -rf public/apipie-cache/* $GITDIR/theforeman.org/plugins/katello/$VERSION/api
       find $GITDIR/theforeman.org/plugins/katello/$VERSION/api -name "*.json" -type f -delete
       sed -i "/layout: /d" $GITDIR/theforeman.org/plugins/katello/$VERSION/api/index.md
+- [ ] Prepare "Katello Next" and future redmine versions
+  - [ ] Rename the "Katello Next" release to be the newest Katello x.y.0 release (e.g. Katello 4.5.0)
+  - [ ] Recreate the "Katello Next" release and indicate that it is a placeholder for issues belonging to the next version of Katello
+  - [ ] Create the next 'z' release of the latest branched Katello version (e.g. Katello 4.4.1)
+  - [ ] Create the next "Katello x.y.z TODO" custom query using the last one as a template
```
jlsherrill, line 97: is there an example of this TODO query?   i'm not familiar with it, so right off i wouldn't know what to do here.

ianballou: I have the first template one created: https://projects.theforeman.org/projects/katello/issues?query_id=170

ianballou: Do you think it's enough to always have one available for reference under the custom queries?

jlsherrill: ahh, i think thats fine!

jlsherrill (approved):

procedures/katello/branch.md.erb
ekohl: What we typically do in Foreman is to have 3 queries. Right now we have 3.0.2, 3.1.2 and 3.2.0. Those are the possible next releases. Once we release 3.1.2, we modify the query to 3.1.3. Once 3.2.0 is released, 3.0.2 is modified to 3.3.0 (since 3.0 is then end of life) and 3.2.0 to 3.2.1.

I know we haven't written this down, but we should. Would this be a good thing to also have for Katello?

procedures/katello/branch.md.erb
```diff
@@ -90,6 +90,12 @@
       cp -rf public/apipie-cache/* $GITDIR/theforeman.org/plugins/katello/$VERSION/api
       find $GITDIR/theforeman.org/plugins/katello/$VERSION/api -name "*.json" -type f -delete
       sed -i "/layout: /d" $GITDIR/theforeman.org/plugins/katello/$VERSION/api/index.md
+- [ ] Prepare "Katello Next" and future redmine versions
+  - [ ] Rename the "Katello Next" release to be the newest Katello x.y.0 release (e.g. Katello 4.5.0)
```
ekohl, line 94: We have a variable for this:
```suggestion
  - [ ] Rename the "Katello Next" release to be the newest Katello <%= develop %>.0 release
```

It's defined here:
https://github.com/theforeman/tool_belt/blob/00e5cc21bad5c4db06acd4f1066e01373cd7fa77/lib/tool_belt/commands/procedure.rb#L21

ianballou:
@ekohl I've made some updates to use variables.

ekohl (approved):
Thanks!

PR #437, opened 2022-03-16
Adding Config for Pulpcore 3.18

Odilhao:


configs/pulpcore/3.18.yaml
```diff
@@ -0,0 +1,78 @@
...
+    based_off: pulpcore-3.17-el8
+    helper_tags:
+      pulpcore-3.18-el8-override: null
+    build_target: pulpcore-3.18-el8-build
+    build_package_group_source_tag: pulpcore-3.18-el8-build
+    arches:
+      - x86_64
+    inherits:
+      pulpcore-3.18-el8-build:
+        pulpcore-3.18-el8-override: 0
```
zjhuntin, line 24: Any reason not to inherit from pulpcore-3.17?

Odilhao: Looking at the configs for 3.17 and 3.16, it does inherit the Current version not the previous one. This is probably the inheritance to compose the packages.

zjhuntin (approved):

PR #450, opened 2022-05-11
Fix based_off for foreman 3.3

Odilhao:
I can't see what changed here

```
    based_off: foreman-plugins-nigthly-nonscl-rhel7
    based_off: foreman-plugins-nightly-nonscl-rhel7
```

But I had to change this


evgeni:
nigthly has a weird order of G, T and H, while nightly does not? ;)

Odilhao:
> nigthly has a weird order of G, T and H, while nightly does not? ;)

I think I need a screen reader

PR #472, opened 2022-08-02
Add 4.6 release config

sjha4:
Adding 4.6 katello config.

### DIFF
--- a/configs/katello/4.6.yaml
+++ b/configs/katello/4.6.yaml
@@ -1,0 +1,110 @@
+---
+:project: katello
+:github_org: katello
+:releases:
+  :4.6:
+    :redmine_version_id: 1561
+:prior_releases:
+  :4.5:
+    :redmine_version_id: 1543
+:code_name: Belgian Tripel
+:mash_scripts:
+  - katello-4.6-el8
+  - katello-candlepin-4.6-el8
+:repos:
+  :katello:
+    :main: true
+    :branch: KATELLO-4.6
+    :repo: https://github.com/Katello/katello.git
+    :version_branch: true
+  :foreman-packaging:
+    :branch: rpm/3.4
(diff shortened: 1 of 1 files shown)


configs/katello/4.6.yaml
```diff
@@ -0,0 +1,142 @@
...
+    :repo: https://github.com/theforeman/foreman_virt_who_configure.git
+  :smart_proxy_container_gateway:
+    :branch: 1.0.6
+    :repo: https://github.com/Katello/smart_proxy_container_gateway.git
+  :katello-selinux:
+    :branch: 4.0.2
+    :repo: https://github.com/Katello/katello-selinux.git
+:ignores:
+:strict_keys: true
+:gpg_key: 2A6A4BAA
```
sjha4, line 45: Needs to be updated.

configs/katello/4.6.yaml
```diff
@@ -0,0 +1,142 @@
...
+:tags:
+  - name: katello-4.6-thirdparty-candlepin-rhel7
+    based_off: katello-thirdparty-candlepin-rhel7
+    build_target: katello-4.6-rhel7-build
+    arches:
```
sjha4, line 50: Do we have to remove the el7 config?

ekohl: Yes, also for nightly though not necessarily in this PR.

configs/katello/4.6.yaml
ekohl: Normally the release engineer (@pcreech this time) creates the key. There's no reason we can't generate the key earlier so we can move it up in the planning if that makes things easier in the future.

sjha4: Thanks @ewoud and @pcreech : Updated keys.

sjha4:
Update host-tools to 4.2.0

pcreech (approved):

PR #479, opened 2022-09-01
Add python39-rpm-macros to the foreman and plugins tags

ehelms:
Forgot this little part.

### DIFF
--- a/configs/foreman/nightly.yaml
+++ b/configs/foreman/nightly.yaml
@@ -50,7 +50,7 @@
         - make
         - patch
         - platform-python
-        - python3-rpm-macros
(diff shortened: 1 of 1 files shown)


ekohl:
@ehelms what's the status of this?

ehelms:
> @ehelms what's the status of this?

That I think it should be merged and executed. We are using `flat-python39-el8` as an external repository already.

ekohl (approved):
Is it then only nightly or also stable release by now?

evgeni:
will this break building foreman maintain the way it's broken downstream?
(Maintain has Python bits that need system (3.6) Python)

PR #486, opened 2022-10-12
Update Katello 4.4 and 4.5 configs

ianballou:
Adds some versions and some ignores.


ianballou:
@Odilhao I don't have merge permissions.

PR #487, opened 2022-11-01
Add 4.7 release config

chris1984:
@ekohl when you get a chance, can this get a review please?

### DIFF
--- a/configs/katello/4.7.yaml
+++ b/configs/katello/4.7.yaml
@@ -1,0 +1,110 @@
+---
+:project: katello
+:github_org: katello
+:releases:
+  :4.7.0:
+    :redmine_version_id: 1613
+:prior_releases:
+  :4.6.0:
+    :redmine_version_id: 1561
+:code_name: Whiskey Sour
+:mash_scripts:
+  - katello-4.7-el8
+  - katello-candlepin-4.7-el8
+:repos:
+  :katello:
+    :main: true
+    :branch: KATELLO-4.7
+    :repo: https://github.com/Katello/katello.git
+    :version_branch: true
+  :foreman-packaging:
+    :branch: rpm/3.5
+    :repo: https://github.com/theforeman/foreman-packaging.git
+  :hammer-cli-katello:
+    :branch: 1.7.z
+    :repo: https://github.com/Katello/hammer-cli-katello.git
+  :katello-host-tools:
+    :branch: 4.2.2
+    :repo: https://github.com/Katello/katello-host-tools.git
+  :katello-certs-tools:
+    :branch: 2.9.0
+    :repo: https://github.com/Katello/katello-certs-tools.git
+  :foreman_virt_who_configure:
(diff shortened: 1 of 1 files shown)


configs/katello/4.7.yaml
```diff
@@ -0,0 +1,110 @@
+---
+:project: katello
+:github_org: katello
+:releases:
+  :4.7:
+    :redmine_version_id: 1613
```
ekohl, line 6: I did notice that the due date is in the future for this: https://projects.theforeman.org/versions/1613

configs/katello/4.7.yaml
```diff
@@ -0,0 +1,110 @@
+---
+:project: katello
+:github_org: katello
+:releases:
+  :4.7:
```
ekohl, line 5: Should this be 4.7.0?

configs/katello/4.7.yaml
```diff
@@ -0,0 +1,110 @@
...
+    :redmine_version_id: 1613
+:prior_releases:
+  :4.6:
```
ekohl, line 8: And should this be 4.6.0?

configs/katello/4.7.yaml
```diff
@@ -0,0 +1,110 @@
...
+    :redmine_version_id: 1561
```
ekohl, line 9: I noticed the version was still open, but since it was released it should be closed. I now did that but perhaps something is missing in the process.

configs/katello/4.7.yaml
```diff
@@ -0,0 +1,110 @@
...
+    :repo: https://github.com/theforeman/foreman_virt_who_configure.git
+  :smart_proxy_container_gateway:
+    :branch: 1.0.6
+    :repo: https://github.com/Katello/smart_proxy_container_gateway.git
+  :katello-selinux:
+    :branch: 4.0.2
+    :repo: https://github.com/Katello/katello-selinux.git
+:ignores:
+:strict_keys: true
+:gpg_key: F120E13E
```
ekohl, line 43: I just created the key. Please update this
```suggestion
:gpg_key: 8f5ca95b
```

ekohl (changes requested):

configs/katello/4.7.yaml
chris1984: That was leftover from when we made the release, noticed the description said placeholder for katello 4.7, changed the description to just Katello 4.7 and set the target date to be Nov 8th when Foreman branches.

configs/katello/4.7.yaml
chris1984: #
Fixed.

configs/katello/4.7.yaml
chris1984: Fixed

chris1984:
@ekohl this is updated, when you get a chance to look again.

ekohl (approved):

PR #491, opened 2022-11-14
Fix various issues with Katello's procedures

ekohl:
See individual commits for details.

### DIFF
--- a/procedures/foreman/branch.md.erb
+++ b/procedures/foreman/branch.md.erb
@@ -77,7 +77,10 @@
     ```
     scp mash_scripts/foreman/<%= release %>.0/*.mash root@koji.katello.org:/etc/mash/
     ```
-- [ ] Update mash script if needed - [collection-mash-split.py](https://github.com/theforeman/foreman-infra/blob/master/koji/collection-mash-split.py)
+- [ ] Update [collection-mash-split.py](https://github.com/theforeman/foreman-infra/blob/master/koji/collection-mash-split.py)
+  - Add a Katello version mapping
+  - Make any changes if needed (usually happens when there are tag changes)
+  - Deploy using `scp collection-mash-split.py root@koji.katello.org:/usr/local/bin`
 - [ ] Add the new release to the JJB job definitions in [jenkins-jobs](https://github.com/theforeman/jenkins-jobs)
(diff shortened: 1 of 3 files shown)


procedures/foreman/branch.md.erb
```diff
@@ -77,7 +77,7 @@
     ```
     scp mash_scripts/foreman/<%= release %>.0/*.mash root@koji.katello.org:/etc/mash/
     ```
-- [ ] Update mash script if needed - [collection-mash-split.py](https://github.com/theforeman/foreman-infra/blob/master/koji/collection-mash-split.py)
+- [ ] Update [collection-mash-split.py](https://github.com/theforeman/foreman-infra/blob/master/koji/collection-mash-split.py) at least add a Katello version mapping. Deploy using `scp collection-mash-split.py root@koji.katello.org:/usr/local/bin`
```
ehelms, line 80: This wording reads a bit confusing, is this meant:

```suggestion
- [ ] Update [collection-mash-split.py](https://github.com/theforeman/foreman-infra/blob/master/koji/collection-mash-split.py) with a Katello version mapping. Deploy using `scp collection-mash-split.py root@koji.katello.org:/usr/local/bin`
```

ehelms (approved):

procedures/foreman/branch.md.erb
ekohl: What I tried to say is that there must be a katello version mapping. However, it should also be checked if there are other changes needed. It's not often, but when we do add el9 that may need changes. I welcome suggestions on how to make that clear. I also considered splitting it up into multiple items so it's clear there are multiple steps.

ehelms: > What I tried to say is that there must be a katello version mapping. However, it should also be checked if there are other changes needed. It's not often, but when we do add el9 that may need changes. I welcome suggestions on how to make that clear. I also considered splitting it up into multiple items so it's clear there are multiple steps.

I think multiple would help -- identify the thing that must happen and then a line saying also consider yadda yadda

ekohl: How about this updated version?

ehelms (approved):

PR #493, opened 2022-11-23
Fix template syntax & provide ERB filename

ekohl:
If it fails to render, the filename becomes part of the stack trace. That makes it easier to find out where it failed.

Fixes: 5d85fa5e5171da66475efb3b13940432a1813b13

### DIFF
--- a/lib/tool_belt/template.rb
+++ b/lib/tool_belt/template.rb
@@ -3,12 +3,10 @@ require 'ostruct'
 
 module ToolBelt
   module Template
-    def self.render(template, vars)
-      ERB.new(template, trim_mode: '-').result(OpenStruct.new(vars).instance_eval { binding })
-    end
-
     def self.render_file(filename, context)
-      self.render(File.read(filename), context)
+      erb = ERB.new(File.read(filename), trim_mode: '-')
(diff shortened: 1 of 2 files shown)


lib/tool_belt/template.rb
```diff
@@ -3,12 +3,10 @@
...
 module ToolBelt
   module Template
-    def self.render(template, vars)
-      ERB.new(template, trim_mode: '-').result(OpenStruct.new(vars).instance_eval { binding })
-    end
-
     def self.render_file(filename, context)
-      self.render(File.read(filename), context)
+      erb = ERB.new(File.read(filename), trim_mode: '-')
+      erb.filename = File.expand_path(filename) if filename
```
evgeni, line 8: When can `filename` be false-y? It managed to read the file in the line before, so it's pretty sure set?

ekohl: That's a leftover from when I added a `filename` parameter to `render` before. Then I realized this method was the only place that used it so I dropped the `render` method instead. Will drop.

evgeni: :+1:

evgeni (approved):
I like it (pending the `if filename` drop ;))

PR #495, opened 2022-11-29
Update tags instructions

ekohl:
Current obal does this automatically so it no longer needs to happen.

I'm not 100% sure this is true for all releases. At least for Foreman 3.4 this wasn't needed. As for yggdrasil: I'm not quite sure if that's still needed at all.

@ehelms @evgeni mind having a look?


evgeni:
Yeah, I think it is fixed for `foreman-release` by the work @ehelms did around tagging/un-tito-ing.

ygg however is still "listed" only on proxy (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/package_manifest.yaml) but in reality needed for proxy and client.

ehelms:
That is a great question - if a PR is merged that needs to do a build *and* a tagging operation I don't think the work I did will fix that solution. The work I did only handles tags when something is merged in packaging that had a build already in existence in Koji.

evgeni:
I think it does accidentally! With tito, a package that went to multiple target tags was built in all of them in parallel. With your changes, it is not. So now the actions become build server-el8, ....., tag client-el8?

ekohl:
So what's the consensus here? Any further work needed or is it ready to merge?

ekohl:
> As for yggdrasil: I'm not quite sure if that's still needed at all.

It is not. It happens during branching so for RC1 it's not needed. Beyond that it should happen when a new yggdrasil release is made, not part of the release process. Otherwise you don't get it in nightly either.

Updated this PR now based on what I had to do during Foreman 3.6 branching.

ehelms (approved):

evgeni:
Just to bring some closure to my worries:
In https://github.com/theforeman/foreman-packaging/pull/9230 + https://github.com/theforeman/obal/pull/341 + https://github.com/theforeman/foreman-packaging/pull/9259 we indeed came to the conclusion that there is no more "I have to tag stuff that was built elsewhere" steps needed, as `obal` with the `non-tito` backend does the right thing

PR #503, opened 2023-01-23
Add Katello 4.8 and update translation & branch steps

chris1984:
@ekohl here is the 4.8 config with the updated translation steps and version bump that Evgeni wanted.


ekohl (approved):
Thanks!

PR #504, opened 2023-01-25
Adding appstream for rust buildtools on Pulpcore 3.21

Odilhao:


configs/pulpcore/3.21.yaml
```diff
@@ -42,6 +42,8 @@
         mode: bare
       - name: centos8-configmanagement-ansible-29
         mode: bare
+      - name: centos8-appstream
```
pcreech, line 45: This won't work as intended.  We'll need to add a new external repo to koji 'flat-rust-toolset' that points to the flattened module.

pcreech: Something a-la https://github.com/theforeman/foreman-infra/commit/ef9f7cc9d9bb70ea52f4b149dcb648ff14849f5d

but using http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001-mod/rust-toolset:rhel8/  as the external repo

Odilhao: Done with theforeman/foreman-infra#1813

pcreech (approved):

PR #514, opened 2023-03-14
Remove individual checkboxes for tagging and signing

ekohl:
These steps are usually ran rather atomically and indivual checkboxes are too much. It still keeps multiple lines because there are multiple scripts.

### DIFF
--- a/procedures/foreman/branch.md.erb
+++ b/procedures/foreman/branch.md.erb
@@ -10,7 +10,7 @@
 
 ## Installer Maintainer
 
-- [ ] Make releases of installer modules
+- Make releases of installer modules
   - Tier 0 (no dependencies)
     - [ ] [candlepin](https://github.com/theforeman/puppet-candlepin)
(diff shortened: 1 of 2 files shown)


ehelms:
As someone who recently did this process and hit errors in parts of these, I appreciate having the individual check boxes as a way to check off each individual step and remind myself where I am at if I get interrupted.

ekohl:
Should we then go the other direction and remove the checkbox from the top, while leaving the individual ones? So `Tag the projects` and `Sign Tarballs`.

ehelms:
> Should we then go the other direction and remove the checkbox from the top, while leaving the individual ones? So `Tag the projects` and `Sign Tarballs`.

Yes, I usually do find the top level checkbox a bit redundant when there are individual ones.

ekohl:
I didn't touch some project branching since https://github.com/theforeman/tool_belt/pull/516 will rewrite that. For packaging branching I'm working on a rewrite as well where I move Katello packaging branching from Katello's branching procedure to Foreman's procedure.

PR #519, opened 2023-03-15
Add is_first_rc and is_first_ga context

ekohl:
Before this we used a lot of checks on the full version, but very often it was checked if it was the first RC or first GA. All other versions are typically the same. This makes it easier to follow along.

### DIFF
--- a/procedures/foreman/release.md.erb
+++ b/procedures/foreman/release.md.erb
@@ -4,7 +4,7 @@
 
 * Release Owner: <%= owner %>
 * Release Engineer: <%= engineer %>
-<% unless is_rc && full_version.end_with?('1') -%>
+<% unless is_first_rc -%>
 * Installer Maintainer: @
 <% end -%>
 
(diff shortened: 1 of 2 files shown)


procedures/foreman/release.md.erb
```diff
@@ -4,7 +4,7 @@
 
 * Release Owner: <%= owner %>
 * Release Engineer: <%= engineer %>
-<% unless is_rc && full_version.end_with?('1') -%>
```
ekohl, line 7: Technically this is a change, since it previously matched on `rc11` too, but that was a bug which we never hit since we've never done so many RCs.

procedures/foreman/release.md.erb
```diff
@@ -25,25 +25,25 @@
 - [ ] Update [docs.theforeman.org](https://github.com/theforeman/foreman-documentation)
   - Using [redmine_release_notes](https://github.com/theforeman/foreman-documentation/blob/<%= short_version %>/guides/doc-Release_Notes/redmine_release_notes) script (see [README](https://github.com/theforeman/foreman-documentation/blob/<%= short_version %>/guides/doc-Release_Notes/README.md) as well): `./guides/doc-Release_Notes/redmine_release_notes foreman <%= version %> > ./guides/doc-Release_Notes/topics/foreman-<%= version %>.adoc`
   - Append CLI release notes taken from the [hammer-cli](https://github.com/theforeman/hammer-cli/blob/<%= short_version %>-stable/doc/release_notes.md) and [hammer-cli-foreman](https://github.com/theforeman/hammer-cli-foreman/blob/<%= short_version %>-stable/doc/release_notes.md) changelogs to [`foreman-<%= version %>.adoc`](https://github.com/theforeman/foreman-documentation/tree/<%= short_version %>/guides/doc-Release_Notes/topics/foreman-<%= version %>.adoc).
-<% unless is_rc && !full_version.end_with?('1') -%>
+<% if is_first_rc || (!is_rc && !is_first_ga) -%>
```
ekohl, line 28: I wasn't sure if there was a clearer way to write it; this is supposed to be done on RC 1 and .1 release, but not RC 2+ and .0 since then it's already part of the manual.

PR #527, opened 2023-04-06
Rely on a script in theforeman-rel-eng to start tool_belt config

ekohl:
Depends on https://github.com/theforeman/theforeman-rel-eng/pull/214


ekohl:
Now part of https://github.com/theforeman/theforeman-rel-eng/pull/214

PR #530, opened 2023-04-12
Add a standalone procedure command

ekohl:
This only uses Ruby standard libraries, which allows running without bundler. That makes it easier to run from other places with minimal setup.

Still need to verify this, so a draft.


ekohl:
Replaced by https://github.com/theforeman/theforeman-rel-eng/pull/240

PR #531, opened 2023-04-19
Add Katello 4.7.5 to config releases and fix some 4.7 releases

ianballou:
I noticed that 4.7 had its own releases in the prior releases section, so I fixed it.

### DIFF
--- a/configs/katello/4.7.yaml
+++ b/configs/katello/4.7.yaml
@@ -2,17 +2,23 @@
 :project: katello
 :github_org: katello
 :releases:
+  :4.7.5:
+    :redmine_version_id: 1705
   :4.7.4:
-   :redmine_version_id: 1697
-:prior_releases:
+    :redmine_version_id: 1697
   :4.7.3:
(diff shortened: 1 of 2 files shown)


chris1984 (approved):

chris1984:
@zjhuntin can this get a merge?

chris1984 (changes requested):
@ianballou Can you add the 4.7.4 to the prior releases section?

ianballou:
> @ianballou Can you add the 4.7.4 to the prior releases section?

Does 4.7.4 need to be in the prior releases part if it's already in the releases section? I updated it, but it's kinda weird that tool belt wouldn't also search the releases section for other cherry picks.

chris1984:
Ignore my comment, if you don't mind reverting my comment. looking at the other configs those 4.7.x need to be in the releases and 4.6 would be in the prior spot

chris1984 (approved):
ACK, @ekohl can this get a merge?

PR #535, opened 2023-05-17
Move mo-files command usage to Katello release time

ianballou:
The mo-files command used to commit to Git. We're now changing it to only generate the files. In the past, since the mo files are gitignored, that command was a no-op.

I'm also changing the procedure to use tx-update, which runs tx-pull and gives some more information.

Related Katello PR: https://github.com/Katello/katello/pull/10557

### DIFF
--- a/procedures/katello/branch.md.erb
+++ b/procedures/katello/branch.md.erb
@@ -21,7 +21,6 @@
   - [ ] `make -C locale tx-pull` in the katello directory
   - [ ] `grunt i18n:compile` in the katello/engines/bastion_katello directory
   - [ ] `bundle exec rake plugin:gettext[katello]` in the foreman directory
-  - [ ] `make -C locale mo-files` in the katello directory
   - [ ] Open a PR to Katello (no Redmine issue needed)
 
 # Three Weeks Prior to Branch Date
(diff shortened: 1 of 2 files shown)


chris1984 (approved):
Makes sense, LGTM

procedures/katello/branch.md.erb
```diff
@@ -18,10 +18,9 @@
   - [ ] [Install grunt](https://github.com/Katello/katello/blob/master/engines/bastion_katello/README.md)
   - [ ] `grunt i18n:extract` in the katello/engines/bastion_katello directory
   - [ ] `tx pull --minimum-perc 50 --all` in the katello directory
-  - [ ] `make -C locale tx-pull` in the katello directory
+  - [ ] `make -C locale tx-update` in the katello directory
```
ekohl, line 21: Technically the only difference between `tx-pull` and `tx-update` is that `tx-update` also warns you to do a follow up step in the foreman directory. That's listed as a step 2 lines below so I think `tx-pull` is the better choice
```suggestion
  - [ ] `make -C locale tx-pull` in the katello directory
```

ianballou: Sounds good, I put it back to tx-pull

ekohl (approved):

PR #536, opened 2023-05-22
Add Katello 4.9 config

ekohl:
This doesn't add `prior_releases`, but at least allows branching packaging.

@wbclark I decided to open this myself rather than waiting for it. Tomorrow (when I branch) I will need it and last time it not being done at the same time caused problems.

### DIFF
--- a/configs/katello/4.9.yaml
+++ b/configs/katello/4.9.yaml
@@ -1,0 +1,80 @@
+---
+:project: katello
+:github_org: katello
+:nightly: false
+:mash_scripts:
+  - katello-4.9-el8
+  - katello-candlepin-4.9-el8
+:tags:
+  - name: katello-candlepin-4.9-el8
+    based_off: katello-candlepin-nightly-el8
+    build_target: katello-4.9-el8-build
+    arches:
+      - x86_64
+  - name: katello-4.9-el8
+    based_off: katello-nightly-el8
+    helper_tags:
+      katello-thirdparty-el8:
+      katello-4.9-el8-override:
+    build_target: katello-4.9-el8-build
+    build_package_group_source_tag: katello-nightly-el8-build
+    arches:
+      - x86_64
+    inherits:
+      katello-4.9-el8-build:
+        katello-4.9-el8-override: 0
+        foreman-plugins-3.7-el8: 3
+        foreman-3.7-el8: 5
+      katello-4.9-el8-override:
+        katello-4.9-el8: 0
+      katello-4.9-el8:
+        katello-thirdparty-el8: 0
+    external_repos:
+      - flat-ruby27-el8
+      - test-flat-el8
+      - centos8-devel
+    build_groups:
+      build:
+        - bash
+        - bzip2
+        - centos-release
+        - coreutils
+        - cpio
+        - diffutils
+        - findutils
+        - foreman-build
+        - gawk
+        - gcc
+        - gcc-c++
+        - grep
+        - gzip
+        - info
+        - make
+        - patch
+        - platform-python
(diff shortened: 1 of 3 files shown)


configs/katello/4.9.yaml
```diff
@@ -0,0 +1,80 @@
+---
+:project: katello
+:github_org: katello
+:nightly: false
+:mash_scripts:
+  - katello-4.9-el8
+  - katello-candlepin-4.9-el8
+:tags:
+  - name: katello-candlepin-4.9-el8
+    based_off: katello-candlepin-4.9-el8
```
evgeni, line 10: this seems like a circular dep, it should be based off nightly

configs/katello/4.9.yaml
```diff
@@ -0,0 +1,80 @@
...
+    build_target: katello-4.9-el8-build
+    arches:
+      - x86_64
+  - name: katello-4.9-el8
+    based_off: katello-4.9-el8
+    helper_tags:
+      katello-thirdparty-el8:
+      katello-4.9-el8-override:
+    build_target: katello-4.9-el8-build
+    build_package_group_source_tag: katello-4.9-el8-build
```
evgeni, line 20: this should be nightly too

configs/katello/4.9.yaml
```diff
@@ -0,0 +1,80 @@
+---
```
evgeni, line 1: this file is missing `:ignores` and `:code_name`, this won't interfere with branching, but something that will be required for when katello actually does the release prep and generates changelog etc

configs/katello/4.9.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - foreman-build
+        - platform-python
+        - python3-rpm-macros
+        - redhat-rpm-config
+        - rhpkg-simple
+        - rpm-build
+        - shadow-utils
+:strict_keys: true
+:gpg_key: bdcc85ae
+:repos: []
```
evgeni, line 77: this too will need filling out, but later

configs/katello/4.9.yaml
```diff
@@ -0,0 +1,80 @@
...
+  - katello-4.9-el8
+  - katello-candlepin-4.9-el8
+:tags:
+  - name: katello-candlepin-4.9-el8
+    based_off: katello-candlepin-4.9-el8
+    build_target: katello-4.9-el8-build
+    arches:
+      - x86_64
+  - name: katello-4.9-el8
+    based_off: katello-4.9-el8
```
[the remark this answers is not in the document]
ekohl: Interesting this wasn't a problem with Foreman's config. Trying to figure out why it happens.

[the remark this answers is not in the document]
evgeni: I mean, it won't be a *problem*, you'll just clone an empty tag?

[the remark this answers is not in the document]
ekohl: I ended up overwriting `version` with `4.9.0` instead of `4.9`, causing the replacement to break.

configs/katello/4.9.yaml
ekohl: That is a also a bug in 3.7. Fixing.

ekohl: https://github.com/theforeman/theforeman-rel-eng/pull/223 & https://github.com/theforeman/tool_belt/pull/537

configs/katello/4.9.yaml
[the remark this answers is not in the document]
ekohl: Updated (and https://github.com/theforeman/theforeman-rel-eng/pull/214 too).

PR #540, opened 2023-05-24
Update Katello 4.9 config

wbclark:


wbclark:
TODO: update hammer-cli-katello and katello-host-tools with new releases

configs/katello/4.9.yaml
```diff
@@ -74,7 +74,38 @@
         - shadow-utils
 :strict_keys: true
 :gpg_key: bdcc85ae
-:repos: []
+:repos:
```
ekohl, line 77: This is just some idea to keep in mind for the future: if add all the repositories in `nightly.yaml` then on branching you "only" need to add all the various branch values. So not for this PR, but it can make future branching easier.

ekohl (approved):

PR #544, opened 2023-06-23
Add Pulpcore 3.28 Config

Odilhao:


configs/pulpcore/3.28.yaml
```diff
@@ -0,0 +1,138 @@
...
+      - name: flat-swig30-el8
+        mode: bare
+      - name: flat-rust-toolset-el8
+        mode: bare
+      - name: flat-llvm-toolset-el8
+        mode: bare
+      - name: centos8-devel
+        mode: bare
+      - name: centos8-configmanagement-ansible-29
+        mode: bare
```
ekohl, line 48: I thought we'd stop using this now that ansible-core is in EL8.

configs/pulpcore/3.28.yaml
```diff
@@ -0,0 +1,138 @@
...
+        pulpcore-3.28-el8: 0
+    external_repos:
+      - name: flat-base-el8
+        mode: bare
+      - name: flat-python36-el8
+        mode: bare
+      - name: flat-python39-el8
+        mode: bare
+      - name: flat-python39-devel-el8
+        mode: bare
```
ekohl, line 36: Will 3.28 be built with Python 3.9? I thought it would use Python 3.11.

Odilhao: Yes, we will keep on 3.9, no rebuild this time

configs/pulpcore/3.28.yaml
Odilhao: That's correct, we should drop this and use ansible-core.

ekohl (approved):

PR #546, opened 2023-07-20
Update Katello 4.9 config's prior releases

ianballou:


wbclark (approved):

configs/katello/4.9.yaml
```diff
@@ -105,6 +105,12 @@
   :4.9.0:
     :redmine_version_id: 1645
 :prior_releases:
+  :4.8.4:
```
ekohl, line 108: For my understanding, is 4.8.4 a prior release? 4.9.0 was released prior to 4.8.4 (which isn't released yet) so does it matter? Or is it defined as all 4.8 versions?

If so, can't you just query the Redmine API at the time of calling? Just open https://projects.theforeman.org/projects/katello/versions.json and filter the name on `Katello 4.8.*`.

ianballou:
I suppose we could just query for all prior releases, perhaps 2 y-streams
back? And yeah 4.8.4 is a "prior release" since issues could be triaged to
4.8.4 but also need to get into 4.9. Without the prior releases, tool_belt
will miss cherry-picks.

PR #549, opened 2023-07-26
Add a help link to make a post a wiki

ekohl:
There was confusion about the process. Readers were assuming that checking off the checkbox made the post a wiki, but it's only a reminder to do so.


ianballou:
@lfu or @wbclark requesting review

wbclark (approved):
The other half of the issue is that, by default, our Discourse accounts don't actually have the permission to make the post into a wiki.

evgeni (approved):

ekohl:
> The other half of the issue is that, by default, our Discourse accounts don't actually have the permission to make the post into a wiki.

That should be covered by https://github.com/theforeman/theforeman-rel-eng#roles but I'm not sure I went back to every person to verify.

PR #552, opened 2023-08-02
Switch to katello release branch before pushing tags

ianballou:
Some releasers accidentally tagged their local branches rather than the upstream release branch. This PR has copy/paste commands for switching to the release branch on the upstream remote.

Question: some users might already have an upstream remote. Would it be better to make a second one? Something like `upstream-release` ?

### DIFF
--- a/procedures/katello/release.md.erb
+++ b/procedures/katello/release.md.erb
@@ -44,6 +44,9 @@
   - [ ] Commit: `git commit -m "Release <%= full_version %>"`
   - [ ] Ensure that the commit above is the _last_ commit and there are no commits after it. This is the commit that will get tagged. (Rearrange commits with `git rebase -i` if needed.)
 - [ ] Once the PR is merged, perform the following in the Katello release branch (the real one, not your fork):
+  - [ ] Create upstream remote: `git remote add upstream https://github.com/Katello/katello.git`
+  - [ ] Fetch upstream remote: `git fetch upstream`
+  - [ ] Checkout upstream release branch: `git checkout upstream/KATELLO-<%= short_version %>`
   - [ ] Tag: `git tag -s -m "Release <%= full_version %>" <%= full_version %>`
   - [ ] Push: `git push --follow-tags` (Must be pushed directly to the release branch, as pull request merges will not preserve tags.)
   - [ ] Generate .mo translation files: `make -C locale all-mo` in the katello directory


procedures/katello/release.md.erb
```diff
@@ -44,6 +44,9 @@
   - [ ] Commit: `git commit -m "Release <%= full_version %>"`
   - [ ] Ensure that the commit above is the _last_ commit and there are no commits after it. This is the commit that will get tagged. (Rearrange commits with `git rebase -i` if needed.)
 - [ ] Once the PR is merged, perform the following in the Katello release branch (the real one, not your fork):
+  - [ ] Create upstream remote: `git remote add upstream https://github.com/Katello/katello.git`
+  - [ ] Fetch upstream remote: `git fetch upstream`
+  - [ ] Checkout upstream release branch: `git checkout upstream/KATELLO-<%= short_version %>`
   - [ ] Tag: `git tag -s -m "Release <%= full_version %>" <%= full_version %>`
```
ekohl, line 50: This is what https://github.com/theforeman/theforeman-rel-eng/blob/master/tag_project already mostly does. Any thoughts on enhancing that to deal with Katello logic? You could also add the gem building process.

procedures/katello/release.md.erb
```diff
@@ -44,6 +44,9 @@
...
   - [ ] Push: `git push --follow-tags` (Must be pushed directly to the release branch, as pull request merges will not preserve tags.)
```
ekohl, line 51: This is what https://github.com/theforeman/theforeman-rel-eng/blob/master/tag_push does.

ekohl:
And unrelated to this, but I just noticed it: the documentation step won't work: it submits a cherry pick PR after the release notes were already made, but the release notes script relies on Fixed in releases which is set by the bot after fixes are merged.

procedures/katello/release.md.erb
ianballou: Sounds like a good idea, how would that work? We would need to set up theforeman-rel-eng with Katello in the projects list, and then have some switches to get the remote and branch naming right?

ekohl: Here's the logic to determine which checkouts to maintain:
https://github.com/theforeman/theforeman-rel-eng/blob/00bc4a681f2893a3d31c3ba08c35c3602741ae3c/ensure_clean_git_checkouts#L5-L13
You can set some variable for all Katello projects here:
https://github.com/theforeman/theforeman-rel-eng/blob/master/releases/katello/settings

The `TAR_PROJECTS` variable is only used in `*_tarballs` scripts which aren't used in Katello's workflow, so you can safely use that.

Then you can check it with `./ensure_clean_git_checkouts`, which should be safe.

ekohl: Some initial work is in https://github.com/theforeman/theforeman-rel-eng/pull/224. While it's focused on branching, it also provides variables that should mostly make it work. I have not tried it and you may need additional scripts to be called.

ekohl:
The procedures have been moved to theforeman-rel-eng, which allows changing scripts and procedures in a single commit. Please open a PR over there now.

ianballou:
https://github.com/theforeman/theforeman-rel-eng/pull/244

PR #555, opened 2023-08-08
Add Katello 4.10

chris1984:

PR #558, opened 2023-08-21
Add katello 4.10 config

ianballou:
Adds on to https://github.com/theforeman/tool_belt/pull/555


pcreech (approved):
ACK, based on what I can see here.  We can add gpg_key and related post-foreman branching items after the fact

PR #562, opened 2023-09-06
Update ignores for katello 4.9.yaml

wbclark:


configs/katello/4.9.yaml
```diff
@@ -125,3 +125,4 @@
   36540:
     # different commits were used for master and 4.9 branches
     - ee53d52ff09b17f1204a88ef7f99207b1d332411
+  36505:
```
ianballou, line 128: Does not including commits ignore every commit for the redmine?

wbclark: Yep, that's the idea: https://github.com/theforeman/tool_belt/commit/8f9b06fa305c7d8e8c5277d2f17f48c8535a4d63

ianballou:
Sorry, my ACK is only grey xD

PR #573, opened 2024-02-08
Katello 4.12 config addition, remove Koji tags and mash scripts from nightly

qcjames53:
This is the config file for Katello 4.12 tooling plus the removal of koji and mash scripts from the nightly config moving forward. Please take care when reviewing this; this is my first branching so there is a good chance there are multiple mistakes in here :)

## Notable questions:
- I updated `flat-nodejs-14-el8` to `flat-nodejs-16-el8`
- Is `centos8-devel` still ok? I understand there have been issues with centos8 recently.


ianballou:
I would recommend looking at both the Katello 4.11 config and nightly to figure out how it should look. I think some sections are missing, like the one that lists all relevant repos, or the one that lists all releases.

What I typically do is make the file match up with the nightly config and replace nightly keywords with "4.12". Then, check the prior release for missing sections.

Your release engineer can help point out anything that is missing, so feel free to ping them too. 

Also, not a hard requirement, but look back to an older one like Katello 4.3... we used to give our releases fun code names but stopped at some point. I miss that a little :)

ianballou:
Also, I totally agree that our documentation here is sorely lacking. The configuration is always been a messy part of branching. If you're interested it could be a great place to improve our branching guide.

chris1984 (commented):
We want to add the sections that are in this file to your PR for tool_belt to pick up what cherrypicks needs to be made:

https://github.com/theforeman/tool_belt/blob/master/configs/katello/4.11.yaml#L9-L44

foreman-packaging goes to 3.10
hammer-cli-katello goes to 1.12.z
smart proxy container gateway goes to 1.2.0
katello-selinux goes to 5.0.3
virt-who-configure goes to 0.5.20
cert tools and host tools stays the same
change the katello branch to 4.12

qcjames53:
Just force pushed some new changes. Thanks for the help, Chris and Ian!

FYI I still left `build_package_group_source_tag` on nightly.

chris1984 (commented):
Looks good on the 4.12 file, did you mean to push up the 4.11 file too?

qcjames53:
I did not. A little confused why there is no diff. Let me take a look.

qcjames53:
It is fixed. I needed to rebase again.

configs/katello/4.12.yaml
```diff
@@ -0,0 +1,116 @@
+---
+:project: katello
+:github_org: katello
+:nightly: false
+
```
chris1984, line 5: Lets remove this space

configs/katello/4.12.yaml
```diff
@@ -0,0 +1,116 @@
...
+  :4.11.0:
+    :redmine_version_id: 1732
+  :4.10.1:
+    :redmine_version_id: 1774
+
+:code_name: Mile 42
+:mash_scripts:
+  - katello-4.12-el8
+  - katello-candlepin-4.12-el8
+
```
chris1984, line 19: Lets remove this space

configs/katello/4.12.yaml
```diff
@@ -0,0 +1,116 @@
...
+  :foreman_virt_who_configure:
+    :branch: 0.5.20
+    :repo: https://github.com/theforeman/foreman_virt_who_configure.git
+  :smart_proxy_container_gateway:
+    :branch: 1.2.0
+    :repo: https://github.com/Katello/smart_proxy_container_gateway.git
+  :katello-selinux:
+    :branch: 5.0.2
+    :repo: https://github.com/Katello/katello-selinux.git
+
```
chris1984, line 48: Lets remove this space

chris1984 (changes requested):
few small comments then it's good to go

configs/katello/4.12.yaml
```diff
@@ -0,0 +1,116 @@
...
+  :4.12.0:
+    :redmine_version_id: 1760
+:prior-releases:
+  :4.11.0:
+    :redmine_version_id: 1732
+  :4.10.1:
+    :redmine_version_id: 1774
+
+:code_name: Mile 42
+:mash_scripts:
```
ekohl, line 16: Like below, release engineering no longer uses this.

configs/katello/4.12.yaml
```diff
@@ -0,0 +1,116 @@
...
+    :branch: 0.5.20
+    :repo: https://github.com/theforeman/foreman_virt_who_configure.git
+  :smart_proxy_container_gateway:
+    :branch: 1.2.0
+    :repo: https://github.com/Katello/smart_proxy_container_gateway.git
+  :katello-selinux:
+    :branch: 5.0.2
+    :repo: https://github.com/Katello/katello-selinux.git
+
+:tags:
```
ekohl, line 49: We're using COPR instead of Koji (since Katello 4.11) so this whole section is no longer used. I don't know if tool_belt can deal with this, but if it can't, the code should be modified to do so.

configs/katello/4.12.yaml
qcjames53: Are you saying the entire :mash_scripts: section can be removed without complication?

configs/katello/4.12.yaml
qcjames53: Ok, thank you for the review Ewoud. Everything in the tags section can be left out? Would it make sense to remove the tags section from the nightly file as well, in that case?

ekohl: Yes, it would. It may be that some tools within tool_belt still demand it somehow. If so, those tools need to be updated. With https://github.com/theforeman/tool_belt/pull/574 I've made a start.

configs/katello/4.12.yaml
qcjames53: Removed

chris1984 (approved):
ACK, the latest change looks good to me, I would adjust the PR title and commit message since we are changing the nightly and 4.11 config as well as 4.12 so we don't have confusion if we have to look back at this PR. 

Will let @ekohl give the last word and look since I don't have merge rights

chris1984 (changes requested):
Actually overlooked the 4.11 i see mash-scripts and tags still in that one

qcjames53:
@chris1984 Are you ok with the PR title as it stands now (only 4.12 and nightly changed)?

ekohl:
@qcjames53 regarding the commit message: [imperative mood](https://cbea.ms/git-commit/#imperative) is the proper tense for the summary (first line).

qcjames53:
Done. Thank you @ekohl :)

configs/katello/nightly.yaml
```diff
@@ -1,75 +1,4 @@
...
-        - bash
-        - centos-release
-        - foreman-build
-        - platform-python
-        - python3-rpm-macros
-        - redhat-rpm-config
-        - rhpkg-simple
-        - rpm-build
-        - shadow-utils
+:nightly: true
```
ekohl, line 4: Please keep the line ending

configs/katello/4.12.yaml
```diff
@@ -0,0 +1,41 @@
...
+    :repo: https://github.com/Katello/katello-certs-tools.git
+  :foreman_virt_who_configure:
+    :branch: 0.5.20
+    :repo: https://github.com/theforeman/foreman_virt_who_configure.git
+  :smart_proxy_container_gateway:
+    :branch: 1.2.0
+    :repo: https://github.com/Katello/smart_proxy_container_gateway.git
+  :katello-selinux:
+    :branch: 5.0.2
+    :repo: https://github.com/Katello/katello-selinux.git
```
ekohl, line 41: Please add a line ending.

qcjames53:
Added the endings you requested, thanks :)

ekohl (approved):
From a technical point of view :+1: and I started the tests. I didn't check the version numbers.

chris1984 (approved):
ACK, looks good from a version standpoint based on the comments I left in the PR discussion and the file looks good now with the stuff gone that is no longer needed. @ekohl can you merge it?

PR #574, opened 2024-02-16
Drop Koji related code and BranchDocsCommand

ekohl:
Since Foreman 3.9 we no longer use Koji but instead use COPR. That means the tags and mash scripts code can be dropped.

Since Katello 4.0 the documentation lives in foreman-documentation so this is no longer needed.


evgeni:
Can we wait until 3.8 is EOL, so if we ever have to reconfigure that, we can?

ekohl:
Yes we can.

PR #577, opened 2024-08-20
Add config for Katello 4.14

chris1984:
@pcreech here is the 4.14 config

### DIFF
--- a/configs/katello/4.14.yaml
+++ b/configs/katello/4.14.yaml
@@ -1,0 +1,43 @@
+---
+:project: katello
+:github_org: katello
+:nightly: false
+:releases:
+  :4.14.0:
+    :redmine_version_id: 1808
+:prior-releases:
+  :4.13.1:
+    :redmine_version_id: 1857
+  :4.13.0:
+    :redmine_version_id: 1788
+  :4.12.2:
+    :redmine_version_id: 1856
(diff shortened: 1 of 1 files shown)


ekohl:
@chris1984 release engineering doesn't use tool_belt at all anymore (since Foreman 3.10). AFAIK these days it's only for the release owner to do cherry picks.

chris1984:
> @chris1984 release engineering doesn't use tool_belt at all anymore (since Foreman 3.10). AFAIK these days it's only for the release owner to do cherry picks.

@ekohl I was just tagging him since he was the release engineer to get it merged. You are right it's only used for cherry-picks now. I was getting a PR based on the branching docs:

```bash
Add tool_belt config for new release:
Create a new yaml file using the nightly Katello config as a template: [tool_belt configs](https://github.com/theforeman/tool_belt/tree/master/configs/katello)
Manually update the following sections:
releases: update to current release. Move the previous ‘current’ release to prior_releases below.
prior_releases: Remove the oldest prior_release (check with that release owner first to see if there’s a reason it should stay)
repos: Update branch names to current versions, including any new releases that need to happen
ignores: Delete all items from this list and start fresh (this will be used for cherry-picks later)
tags: update Katello version number in all values. Check the nightly config to see if any tags/repos need to be updated
```

